PR #2, opened 2018-11-08
Add kafka-test module

garyrussell:


snicoll:
Looks like #9 would be more in line with what we're doing atm.

garyrussell:
@snicoll Sure, no problem; I simply based this on something else in the file (perhaps the embedded mongo).

I just find it a pain to have to manually add the dep every time I create a project.

PR #14, opened 2018-11-19
Upgrade to Spring Cloud Greenwich.M3

spencergibb:
We had to do a quick release for some missing merges, so in effect, there is no M2.


snicoll:
Thanks for the feedback.

spencergibb:
And the thank you for all the work you put into initializr!

PR #21, opened 2018-11-29
Upgrade to Apache Camel 2.23.0

davsclaus:
Signed-off-by: Claus Ibsen <claus.ibsen@gmail.com>


snicoll:
@davsclaus thanks! Is that release compatible with Spring Boot 2.1? Apache Camel is disabled at the moment with the current version.

davsclaus:
Should we have a version range for SB 2.0.x for Camel 2.22.x and older, and then only SB 2.1 for Camel 2.23 onwards ?

snicoll:
That's the plan, yes. Thanks for confirming this release supports SB 2.1 now. I'll polish the PR right away.

PR #34, opened 2018-12-24
Clarify SCS client library dependencies

royclarkson:
@csterwa, @snicoll requested we clarify that these are client libraries. I've made the obvious change, but if you would prefer it be something different, then please let us know.


mbhave:
Thanks @royclarkson. Maybe the description should be updated to make it more explicit that these are clients for use with the Spring Cloud Services on PCF?

snicoll:
@royclarkson gentle ping

snicoll:
Closing due to lack of feedback. See also #79

PR #44, opened 2019-01-23
Skip YAML binding for SSL integration test that doesn't need it

dsyer:
Shaves about 1.5sec off the test suite (out of 13).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -27,6 +27,12 @@
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-web</artifactId>
+			<exclusions>
+				<exclusion>
+					<groupId>org.hibernate.validator</groupId>
+					<artifactId>hibernate-validator</artifactId>
+				</exclusion>
+			</exclusions>
 		</dependency>
 
 		<dependency>
@@ -45,12 +51,6 @@
 		<dependency>
 			<groupId>org.ehcache</groupId>
 			<artifactId>ehcache</artifactId>
-		</dependency>
-
-		<dependency>
-			<groupId>org.springframework.boot</groupId>
-			<artifactId>spring-boot-devtools</artifactId>
-			<optional>true</optional>
 		</dependency>
 
 		<dependency>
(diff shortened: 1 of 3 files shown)


pom.xml
```diff
@@ -47,12 +53,6 @@
 			<artifactId>ehcache</artifactId>
 		</dependency>
 
-		<dependency>
```
snicoll: Did you intend to remove devtools?

src/test/java/io/spring/start/site/StartApplicationHttpsTests.java
```diff
@@ -41,7 +41,9 @@
  * @author Stephane Nicoll
  */
 @RunWith(SpringJUnit4ClassRunner.class)
-@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, properties = "initializr.env.force-ssl=true")
+@SpringBootTest(webEnvironment = WebEnvironment.RANDOM_PORT, properties = {
+		"initializr.env.force-ssl=true",
+		"spring.config.location=file:./src/test/resources/application-test.yml" })
```
snicoll: This means that tests have to be ran from the module directory. This sounds kind of obvious but I don't like the idea of hardcode relative paths here.

src/test/resources/application-test.yml
```diff
@@ -0,0 +1,28 @@
+local:
```
snicoll: I don't understand the content of this file. The `gcp.version` seems unrelated to the task at hand. Same thing for logging & co.

pom.xml
dsyer: No, I guess not. Sorry.

src/test/java/io/spring/start/site/StartApplicationHttpsTests.java
dsyer: Where else would they be run from? We can use `classpath:...` I suppose. Want me to change it?

src/test/resources/application-test.yml
dsyer: It's just the `application.yml` without the `initializr.*` properties. There didn't seem a lot of point fretting about the rest, since they are pretty fast to bind.

snicoll: If we have a dedicated configuration file for a test (that's not a profile) I'd expect:

1. To only contain what is required
2. To not be named like it's a standard file that would be picked up if a `test` profile is active.

snicoll:
Thanks for the PR. In the meantime, we've added a full integration tests suite that validates a lot more variables but also increases significantly the total build time. Given that 2.2 will improve this, the extra complexity in the build setup is not justified IMO.

PR #61, opened 2019-02-21
Improve Spring Cloud Task Description

mminella:
Spring Cloud Task provides result tracking of finite tasks as well as
integration with Spring Batch and Spring Cloud Stream.  This commit adds
the detail of integrating with Spring Cloud Stream to the description.


snicoll:
Thank you, this will be available in production shortly.

PR #66, opened 2019-03-06
Upgrade to Vaadin 13.0.0

qtdzz:


snicoll:
@qtdzz thanks but I am not keen to jump to the next feature release like that. Perhaps we should consider 13 for 2.2 only? Paging @mstahv for feedback.

qtdzz:
@snicoll Thanks for your comment. I agree! I will ping @mstahv

mstahv:
I think we can go to 13.0.0, as long as somebody has made proper testing for it. Currently I suggest all 10+ users to go with the latest, not with LTS. 

11,12 and 13 have basically been minor releases (with important backwards compatible features that existed in V8 and earlier), but due to the LTS model we started to use the major version has been upgraded as well. Hopefully from next LTS (14,15 or something), we could stick in the latest bugfix release of LTS version.

If 13.0.0 is in Maven central already, I could do some additional smoke tests before landing this.

mstahv:
Tested with on "production app" that now runs on Spring Boot 2.1.3.RELEASE and Vaadin 13.0.0 and did a smoke test with Spring initializer (manually updated version to 13.0.0). All seems fine! Let's land it, 13 has some very important features and fixes (my "production app" was using 13.0.0.beta versions before).

snicoll:
Roger

Issue #79, opened 2019-03-16
Review entries metadata


snicoll:
To provide a consistent experience in the UI, a review of the current entries is in order. A large number of entries do not have links at the moment and we could improve the developer experience with them via `HELP.md`.

snicoll:
I think we can close this now.

PR #80, opened 2019-03-19
Clarify SCS client library dependencies

royclarkson:
New PR to address #34 and #79


mbhave:
Thanks @royclarkson. This is now on master.

PR #83, opened 2019-03-22
Update Readme

oodamien:
Revisit the page
Convert to MD


snicoll:
@oodamien the output is much better that way indeed but we're not keen to move to markdown for that reason alone and prefer the consistency of using asciidoc everywhere. Thanks for the PR, in any case!

PR #87, opened 2019-04-01
Upgrade to Vaadin 13.0.3

tulioag:


snicoll:
Thanks for the PR, I've applied that as part of a polish in #84

PR #95, opened 2019-04-07
Upgrade to mybatis spring 2.0.1 and 1.3.4

kazuki43zoo:
The mybatis-spring-boot 2.0.1 and 1.3.4 has been released.


snicoll:
@kazuki43zoo thanks for the update, this will be available shortly.

PR #105, opened 2019-04-29
Add Spring Boot Admin 2.2.0-SNAPSHOT for Spring Boot 2.2.x

joshiste:


snicoll:
I am not keen to add a snapshot of a third party entry unless that's absolutely necessary. If you have milestones, we can add them for sure or wait for GA.

Is there a specific reason you are requesting this?

joshiste:
> Is there a specific reason you are requesting this?

- We had this in the past.
- Generating a project with Spring Boot 2.2.0.M2 and the latest Spring Boot Admin release (currently 2.1.4) results in incompatible dependencies.

snicoll:
> We had this in the past.

Not sure how that makes it a reason. This was different anyway: it was the main line and a one time measure as it was broken due to Spring Cloud related feature if I remember correctly.

> Generating a project with Spring Boot 2.2.0.M2 and the latest Spring Boot Admin release (currently 2.1.4) results in incompatible dependencies.

Cool, thanks for letting us know. I've [updated the version range](https://github.com/spring-io/start.spring.io/commit/eac4455c3af70066f64796614039bf9f9c09085b) to disable Spring Boot admin with `2.2.0` and later. We can restore once you have a milestone/release.

joshiste:
> We can restore once you have a milestone/release.

Hmm we currently don't do any milestone releases - as we are lacking a repository for this (sonatype only has the snapshot repository - and I'm not keen on running my own). So we can't provide working software for the early adopters via start.spring.io

> I am not keen to add a snapshot of a third party entry
What is the problem with third party snapshots on start.spring.io?

PR #106, opened 2019-04-29
Add Spring Cloud Kubernetes

joshiste:


snicoll:
@joshite Thanks for the PR but we do not accept new entries that way as we have a dedicated process via the creation of an issue.

Issue #28, opened 2018-12-12
Mark Spring Cloud Netflix projects as maintenance mode


OlgaMaciaszek:
Information about the tools being in maintenance mode should be added to the `help.md` file.

mbhave:
In addition to this, I think we should disable these dependencies for Boot 2.2.x projects (Spring Cloud Hoxton). I also wonder if we should go a step further and disable them for Boot 2.1.x (Spring Cloud Greenwich).

spencergibb:
No, not for Greenwich. We need to discuss for Hoxton.

snicoll:
Sounds good, let us know what you decide and we can adapt.

@OlgaMaciaszek would you be willing to contribute a PR to add a blurb in `HELP.md`? We're going to roll-out the feature soon and we can include this one as well.

OlgaMaciaszek:
Yes, @snicoll . Will do that.

OlgaMaciaszek:
Submitted a PR, although, later on I think it should be considered to enable adding this warning section on top of `HELP.md` (ordering is currently not supported by Initializr).

snicoll:
Closing in favour of PR #112

PR #112, opened 2019-04-30
Add warning to HELP.md when user requested maintenance mode dependency

OlgaMaciaszek:
Fixes gh-28


mbhave:
Thanks for the PR @OlgaMaciaszek. I think we should also add the recommended replacement for the maintenance mode dependency instead of just saying `We do not recommend adding them to
new projects`

start-site/src/main/resources/templates/maintenance-mode-dependencies.mustache
```diff
@@ -0,0 +1,6 @@
+The dependencies listed below are in maintenance mode. We do not recommend adding them to
```
snicoll: I'd start with something tailored for the spring cloud use cases with a general section that titles exactly what happens and then some blurp that summarizes the situation. Perhaps a link to blog post announcement?

OlgaMaciaszek:
Hi, @snicoll @mbhave thanks for the review. Following your suggestions, I have changed it into something specific to the Spring Cloud Netflix maintenance mode situation. I have also added link to the blog entry that has more details on the background of the maintenance mode decision, as well as the list of suggested replacement. Let me know if it's fine now.

snicoll:
Thanks @OlgaMaciaszek - this is now merged with a [polish commit](https://github.com/spring-io/start.spring.io/commit/1d0c874933ff9dc82c31d50ab93fcbfaa8e0b680).

PR #114, opened 2019-05-03
Upgrade to Spring Cloud Open Service Broker 3.0.0.RELEASE

royclarkson:


snicoll:
@royclarkson thanks for the PR but I prefer to keep the history for `2.x` rather than upgrade to `2.1.x`. We have plans to offer upgrade assistance in the future so having the metadata about older versions is quite interesting. 

I've upgraded to 3.0.0 and 2.1.2 in a separate commit, see 9b834cd

PR #118, opened 2019-05-10
Upgrade to Spring Cloud Services 2.1.3.RELEASE

habuma:


habuma:
@snicoll Just curious...how long before I should expect this to be at start.spring.io? In the past, it has usually been right away, but it's now 2 hours later and the pom.xml file I get from start.spring.io still shows Spring Cloud Services at 2.1.2.RELEASE. No rush...just need to set expectations.

snicoll:
5 minutes.

The CI build that should push those bits to production broke and I am in a call right now and didn't have the time to investigate why. Nothing related to the change.

snicoll:
@habuma should be good now.

PR #125, opened 2019-05-17
Upgrade to Okta starter 1.2.0

bdemers:


snicoll:
Thanks for the update @bdemers. Do you have compatibility build against Spring Boot 2.2? We'd like to make sure this is covered.

bdemers:
I’ll get on that this week!

bdemers:
@snicoll I found a couple issues, one in spring-sec and this one: spring-projects/spring-boot#17035

snicoll:
@bdemers thanks for the feedback and trying but what does that mean concretely? Is it broken or is the spring security a corner case issue?

bdemers:
Sorry @snicoll, I wasn’t clear about the spring sec issuer, that one is a potential vuln, so it is private (and unrelated to the linked one above)

snicoll:
What link? Can you please confirm that the current generation works with SB 2.2? If not, we need to disable it here.

bdemers:
@snicoll fixed in #144

snicoll:
Thanks @bdemers

PR #145, opened 2019-06-06
Fix NPE when Spring Boot metadata URL cannot be accessed

maverick1601:


snicoll:
@maverick1601 thanks for reporting the issue. I've decided to fix this slightly differently and add a test to cover that use case. I've pushed https://github.com/spring-io/start.spring.io/commit/71cb3b5f7d18f250c0206c1bb1e87a1dc1b5c519. Thanks for the PR, in any case!

Issue #149, opened 2019-06-10
Spring Cloud Dependency management plugin is not being set up for the Spring Cloud Tools group


OlgaMaciaszek:
The build files are not set up correctly as the Spring Cloud bom is missing in Spring Cloud Tools group's configuration.

wilkinsona:
Thanks, @OlgaMaciaszek. If you intend to open a pull request there's no need for an issue as well.

OlgaMaciaszek:
Ah, ok. Thanks for the info.

PR #150, opened 2019-06-10
Spring Cloud bom is not imported for dependencies in the Spring Cloud Tools group

OlgaMaciaszek:
Fixes gh-149


OlgaMaciaszek:
Yes, will fix that.

PR #151, opened 2019-06-10
spring-cloud-starter entry is missing

dsyer:
Previously removed in fa647655


snicoll:
We can't hardcode a version in a reference doc link so I've changed it. I've also created https://github.com/spring-cloud/spring-cloud-static/issues/16

PR #154, opened 2019-06-11
Fix gh 29 add scc plugin to build

OlgaMaciaszek:


mbhave:
Thanks @OlgaMaciaszek. It seems like this would be useful as a more general purpose feature supported by the library. 

I'm adding the team-attention label to see what the rest of the team thinks.

start-site/src/main/java/io/spring/start/site/extension/springcloud/SpringCloudProjectGenerationProperties.java
```diff
@@ -0,0 +1,39 @@
...
+package io.spring.start.site.extension.springcloud;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+
+/**
+ * Properties related to generating Spring Cloud projects.
+ *
+ * @author Olga Maciaszek-Sharma
+ */
+@ConfigurationProperties("start.spring-cloud")
```
snicoll: We don't use multi-words in prefixes.

start-site/src/main/java/io/spring/start/site/extension/springcloud/SpringCloudProjectsMetadataProvider.java
```diff
@@ -0,0 +1,34 @@
...
+ */
+public interface SpringCloudProjectsMetadataProvider {
+
+	/**
+	 * Returns Spring Cloud projects metadata for the specified Spring Cloud Release Train
+	 * version.
+	 * @param releaseTrainVersion release Train version as {@link String}
+	 * @return {@link SpringCloudProjectsMetadata} object
+	 */
+	SpringCloudProjectsMetadata get(String releaseTrainVersion);
```
snicoll: That contract is very specific to Spring Cloud and I don't think that's warranted. Rather I would have expected something that would give us a map of version for a given `platformVersion`. 

For instance, I am calling a service for Spring Boot `2.1.5.RELEASE` and I get a bunch of versions, letting the service discover based on what it can discover. We'd get a `spring-cloud-contract` but also a `hibernate` key for the Hibernate version read from the Spring Boot bom.

start-site/src/test/java/io/spring/start/site/extension/springcloud/DefaultSpringCloudProjectsMetadataProviderTests.java
```diff
@@ -0,0 +1,82 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.spring.start.site.extension.springcloud;
+
+import javax.cache.CacheManager;
```
snicoll: We don't use JSR-107 directly but our own abstraction.

start-site/src/main/java/io/spring/start/site/extension/springcloud/DefaultSpringCloudProjectsMetadataProvider.java
```diff
@@ -0,0 +1,81 @@
...
+		implements SpringCloudProjectsMetadataProvider {
+
+	private static final Log LOG = LogFactory
+			.getLog(DefaultSpringCloudProjectsMetadataProvider.class);
+
+	private final RestTemplate restTemplate;
+
+	private final String infoHost;
+
+	DefaultSpringCloudProjectsMetadataProvider(RestTemplate restTemplate,
```
snicoll: Let's use `WebClient` from now on. I'll add an issue to migrate our current use.

wilkinsona:
Thanks, @OlgaMaciaszek. I share Madhura and Stéphane's thoughts about the version resolution being a really useful general purpose feature. We could use it for the Kotlin version of Dependency Management Plugin version for example. To that end, it'd be great to have something that's not tied to Spring Cloud.

We've been wondering about allowing the metadata to specify the coordinates of a bom and an XPath expression that should be used to extract a version from the bom. We'd then grab the bom from Maven Central or repo.spring.io, generate its effective pom, and then extract the value. There's some code in the Dependency Management Plugin that does most of this that we may be able to reuse.

I hope to take a look at doing the above over the next few days. We'd then hopefully be able to merge this PR on top of that new general-purpose functionality.

OlgaMaciaszek:
Hi all, thanks for the review.

@wilkinsona , @snicoll and I have previously talked about creating such a library and extracting versions and other metadata from boms, but then when we talked during Spring.io, we decided we would use cloud-info for cloud projects, since it's already available.

OlgaMaciaszek:
After talking to @mbhave I understand that the approach has changed from what was discussed at Spring I/O, so I will wait till the changes mentioned by @wilkinsona are there and then resume work on this feature.

wilkinsona:
@OlgaMaciaszek I've merged the changes for the new version resolver stuff and we're [now using it](https://github.com/spring-io/start.spring.io/commit/9ee7d7695c346c3eb2df6402914ecbb069e0614a) to get the Kotlin version from `spring-boot-dependencies` rather than hard coding it. I think that means we can now proceed with this PR. Do you have the time to rebase and see if the new API meets your needs?

OlgaMaciaszek:
Hi @wilkinsona Thanks a lot, will do that.

mbhave:
The squashed commits along with a polish commit are [here](https://github.com/mbhave/start.spring.io/commit/7424f3b430f3cd5d192f259e8fc2027a75c27454). I mainly wanted to get everyone's thoughts on changing the API from returning an `Optional` to returning a `String`.

wilkinsona:
We haven't made much use of `Optional` here or in Initializr so +1 for `String` as it feels more consistent.

OlgaMaciaszek:
Changing to `String` is fine with me.

mbhave:
@OlgaMaciaszek Thanks again for the PR! This has now been merged into master along with [this polish commit](https://github.com/spring-io/start.spring.io/commit/24c90cd9bcea9bcb69d7cc8f320d62095da457c6).

PR #166, opened 2019-06-26
Add a test-scoped dependency on com.unboundid:unboundid-ldapsdk when Data LDAP starter is selected

eddumelendez:
Allow to add `com.unboundid:unboundid-ldapsdk` dependency when
spring-ldap is selected.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/SpringLdapBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/SpringLdapBuildCustomizer.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.spring.start.site.extension;
+
+import io.spring.initializr.generator.buildsystem.Build;
+import io.spring.initializr.generator.buildsystem.Dependency;
+import io.spring.initializr.generator.buildsystem.DependencyScope;
+import io.spring.initializr.generator.spring.build.BuildCustomizer;
+
+/**
+ * A {@link BuildCustomizer} that automatically adds "unboundid-ldapsdk" when ldap is
+ * selected.
+ *
+ * @author Eddú Meléndez
+ */
+public class SpringLdapBuildCustomizer implements BuildCustomizer<Build> {
+
+	@Override
+	public void customize(Build build) {
+		if (build.dependencies().has("data-ldap")) {
+			build.dependencies().add("unboundid-ldapsdk", Dependency
+					.withCoordinates("com.unboundid", "unboundid-ldapsdk").scope(DependencyScope.TEST_COMPILE));
+		}
+	}
+
+}
--- a/start-site/src/main/java/io/spring/start/site/extension/StartProjectGenerationConfiguration.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/StartProjectGenerationConfiguration.java
@@ -89,6 +89,11 @@ public class StartProjectGenerationConfiguration {
 	}
 
 	@Bean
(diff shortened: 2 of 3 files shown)


wilkinsona:
Thanks for the PR, @eddumelendez. If we believe that every application that uses Spring Data LDAP should have a dependency on the UnboundID LDAP SDK, should it not be part of `spring-boot-starter-data-ldap`?

eddumelendez:
`com.unboundid:unboundid-ldapsdk` can be used as a embedded ldap for testing purposes. I think the commit message is not the right one.

AFAIK `spring-boot-starter-data-ldap` doesn't provide dependencies with scope `test`.

wilkinsona:
Sorry and thanks for bearing with me, @eddumelendez. I'd completely missed that it was a test dependency.

snicoll:
@eddumelendez I am not entirely sure that we should add this automatically. Wondering what you think of adding a note in `HELP.md` instead.

wilkinsona:
FWIW, I think it probably does make sense to add this automatically. The UnboundID SDK is one of two options described in the [Spring LDAP docs](https://docs.spring.io/spring-ldap/docs/current/reference/#testing). The other option, ApacheDS, isn't really maintained any more.

snicoll:
I am not questioning the choice, I am more questioning the fact that tests are using an embedded LDAP server automatically. This isn't something we haven't done for any other store. We also have an embedded MongoDB option on the site at the moment that feels quite inconsistent with this proposal.

wilkinsona:
I think LDAP's a different case to the other stores. With Mongo, there are a few choices:

1. Embedded Mongo
2. Testcontainers
3. Fongo or the like

With LDAP, UnboundID seems to be the the only widely used choice and sufficiently so to warrant adding it as a test dependency by default.

@jgrandja @jzheaux @christophstrobl  @mp911de what's your take on this please? Is UnboundID SDK sufficiently widely used to test an application that uses Spring (Data) LDAP to warrant adding it as a test-scoped dependency by default?

eddumelendez:
> @eddumelendez I am not entirely sure that we should add this automatically. Wondering what you think of adding a note in `HELP.md` instead.

@snicoll Adding a note in `HELP.md` would be helpful.

But since `unboundid` is the library provided by `spring-boot-dependencies` and also provide auto-configuration I don't see why should not be provided when the project is created. One of the reasons could be that not many projects are using it so that dependency would be remaining but I also see the same approach is applied to `spring-kafka`

snicoll:
> I also see the same approach is applied to spring-kafka

I am not sure It's the same though. Spring Kafka brings test utilities whereas this proposal replaces what the application would normally do (trying to connect to whatever LDAP server is configured). It's not inconsistent with what we do elsewhere though (such as `@AutoconfigureTestDatabase`) though that is explicit rather than relying on classpath).

@jgrandja @jzheaux @christophstrobl @mp911de what's your take on this please?

jzheaux:
> Is UnboundID SDK sufficiently widely used to test an application that uses Spring (Data) LDAP to warrant adding it as a test-scoped dependency by default?

I agree that UnboundID SDK is widely used once the choice is made to use an embedded LDAP server.

> With Mongo, there are a few choices:
> 1. Embedded Mongo
> 2. Testcontainers
> ...

Testcontainers is a legitimate choice as well, likely with similar motives to Mongo.

Given that, would it make more sense to add an "Embedded LDAP server" option that parallels the "Embedded MongoDB database" option?

snicoll:
@eddumelendez I am confused about the PR now. It looks like `spring-ldap-test` brings interesting utilities that should be used in combination with either UnboundID or ApacheDS. Is there a reason why you didn't consider adding it as part of this PR?

A quick look at `spring-ldap-test` I can see it has optional dependencies on them (which is what I expect reading the doc). IMO this PR should be reworked to:

1. Add `spring-ldap-test` when Spring LDAP is selected
2. Add a new entry in the Testing group for unboundid

What do you think?

eddumelendez:
@snicoll `spring-boot` offers integration with unboundid out-of-the-box, we just need the dependency declared. Adding `spring-ldap-test` will let users to choose between Unboundid or ApacheDS for sure but not sure how much value will provide having `spring-ldap-test` and `unboundid` in the classpath if `spring-boot` does all the work already. I understand that if ApacheDS is chosen because `spring-ldap-test` will provide the utilities to perform the testing against that server.

My understanding is that we can provide that flexibility between UnboundId and ApacheDS by having them in the "Testing" group, right?

Something that came to me mind is, by having

1. spring-ldap-test + unboundid, developers have two choices. They can use embedded ldap server auto-configuration provided by spring boot (no value of spring-ldap-test) or use utilities provided by spring-ldap-test.
2. spring-ldap-test + apacheds, developers can use utilities provided by spring-ldap-test

snicoll:
Thank you Eddu, I completely missed our auto-config support. I've gone ahead and added the option in #236 so I am closing this PR. Thanks for your help!

PR #234, opened 2019-08-28
Add option to share the configuration of a project

oodamien:


start-client/src/components/common/error/Warnings.js
```diff
@@ -0,0 +1,90 @@
...
+        if (key === 'dependencies') {
+          warnings.push(
+            `The following dependencies can not be add: <strong>${value.map(
+              dep => dep.value
+            )}</strong>`
+          )
+        }
+        if (key === 'project') {
+          warnings.push(
+            `<strong>${value.value}</strong> is not a valid project type. The default value has been selected.`
```
wilkinsona: I think this would be better if it said what has been selected instead rather than just "The default". In this case that would be  "Maven has been selected" rather than "The default value has been selected".

start-client/src/components/common/error/Warnings.js
```diff
@@ -0,0 +1,90 @@
...
+          )
+        }
+        if (key === 'language') {
+          warnings.push(
+            `<strong>${value.value}</strong> is not a valid language. The default value has been selected.`
```
wilkinsona: Similarly, could this say "Java has been selected" rather than "The default value has been selected"?

start-client/src/components/common/error/Warnings.js
```diff
@@ -0,0 +1,90 @@
...
+          )
+        }
+        if (key === 'meta') {
+          if (get(value, 'java')) {
+            warnings.push(
+              `<strong>${get(
+                value,
+                'java.value'
+              )}</strong> is not a valid Java version. The default value has been selected.`
```
wilkinsona: Similarly, could this say "8 has been selected" rather than "The default value has been selected"?

start-client/src/components/common/error/Warnings.js
```diff
@@ -0,0 +1,90 @@
...
+            )
+          }
+          if (get(value, 'packaging')) {
+            warnings.push(
+              `<strong>${get(
+                value,
+                'packaging.value'
+              )}</strong> is not a valid packaging. The default value has been selected.`
```
wilkinsona: Similarly, could this say "Jar has been selected" rather than "The default value has been selected"?

start-client/src/components/common/error/Boot.js
```diff
@@ -0,0 +1,38 @@
...
+  render = () => {
+    return (
+      <div className='control-error'>
+        <p className='title'>
+          <strong>
+            The version {this.props.value} of Spring Boot is outdated.
+          </strong>
+        </p>
+        <p>
+          You have to select a valid version if you want to generate or explore
```
wilkinsona: How about "Select a valid version to generate or explore the project"?

This also suggests that you can still share a project with invalid configuration but trying to do so results in an error. I guess it should say something like "Select a valid version to generate, explore, or share the project"?

I also wonder if the buttons for generating, exploring, and sharing should change appearance somehow when the configuration is invalid such that they won't do anything.

start-client/src/components/common/error/Boot.js
```diff
@@ -0,0 +1,38 @@
...
+        </p>
+        <p>
+          <a
+            href='/#'
+            onClick={e => {
+              e.preventDefault()
+              this.props.selectLatest()
+            }}
+          >
+            > Select the latest stable version
```
wilkinsona: Could this include what the version is, so something like "Select 2.1.7 (the latest stable version)"?

start-client/src/components/common/share/SharePopup.js
```diff
@@ -0,0 +1,129 @@
...
+  }
+
+  getUrl = () => {
+    const rootValues = this.props.rootValues
+    const values = this.props.properties
+    return getShareUrl(values, rootValues)
+  }
+
+  onCopy = () => {
+    toast.success('Your URL has been copied in the clipboard with success.')
```
wilkinsona: Perhaps just "Project URL has been copied to the clipboard"?

start-client/src/components/common/share/SharePopup.js
```diff
@@ -0,0 +1,129 @@
...
+                    <a
+                      href='/#'
+                      onClick={e => {
+                        e.preventDefault()
+                        this.shareOnTwitter()
+                      }}
+                      className='button button-default'
+                    >
+                      <IconTwitter />
+                      Share on twitter
```
wilkinsona: I don't think we need to provide the ability to share directly to Twitter. IMO, copying the URL to the clipboard is sufficient. It can then be pasted wherever the user wants.

Issue #237, opened 2019-08-31
Make with success messages more concise


wilkinsona:
There are a few actions in the UI that result in a with success message being displayed to those user. Those messages are the following:

- Your file has been copied in the clipboard with success.
- Your file has been downloaded with success.
- Your project has been generated with success
- Your project has been downloaded with success

I think the messages could be tightened up a bit and made more concise without losing any of their meaning. Here's an initial proposal:

- File copied to the clipboard.
- File downloaded.
- Project generated.
- Project downloaded.

The first could perhaps even be "File copied." if there's already enough context to make it clear that the destination is the clipboard. The would align it with the two downloaded messages that do not describe the destination. That would make the messages the following:

- File copied.
- File downloaded.
- Project generated.
- Project downloaded.

snicoll:
I was also wondering if some of these messages could be removed completely. If we invoke the generate project action and, as a result, we download a file, isn't that enough signal for the user?

I've seen myself ignoring the green top header completely as my eye focused on the fact a new file was downloading....

wilkinsona:
It does feel a little strange to display something that tells you that the UI element you just clicked on has worked. On further consideration, I would have no objection to removing all four messages. In fact, I think that's what I would prefer.

oodamien:
We need at least to provide a feedback for the copy to clipboard.
There is no other feedback than this one.

snicoll:
Closing in favour of PR #239

PR #239, opened 2019-09-02
Revisit success messages

oodamien:
Remove success message on project generation
Revisit feedback on file copy
See gh-237

### DIFF
--- a/start-client/src/components/common/explore/ExploreModal.js
+++ b/start-client/src/components/common/explore/ExploreModal.js
@@ -15,16 +15,21 @@ class ExploreModal extends React.Component {
   constructor(props) {
     super(props)
     this.ref = React.createRef()
+    this.state = {
+      button: 'Copy',
+    }
   }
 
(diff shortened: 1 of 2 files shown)


start-client/src/components/common/explore/ExploreModal.js
```diff
@@ -15,16 +15,21 @@ class ExploreModal extends React.Component {
...
     super(props)
     this.ref = React.createRef()
+    this.state = {
+      button: 'Copy',
+    }
   }
 
   onCopy = () => {
-    toast.success('Your file has been copied in the clipboard with success.')
+    this.setState({ button: 'Copied!' })
```
snicoll: Is `Copied!` the message that the user is going to see? Don't we want something more neutral?

snicoll: Nevermind, I thought the message on the guides was a bit different.

PR #249, opened 2019-09-05
Upgrade to Azure Spring 2.1.7

jdubois:


snicoll:
Thanks for letting us know about the new version. It isn't new but I found a problem with dependency management with regards to Spring Boot. I've opened https://github.com/microsoft/azure-spring-boot/issues/716

Issue #240, opened 2019-09-02
start.spring.io/error page doesn't display the dedicate error page


oodamien:
https://start.spring.io/error should display the same page than https://start.spring.io/aaa

snicoll:
closing in favour of PR #260

PR #260, opened 2019-09-10
Add generic error page

oodamien:
See gh-240


snicoll:
Thanks @oodamien but we can't exclude the auto-config as it also deals with non HTML clients. I am looking at various options but nothing satisfactory right now I am afraid.

snicoll:
In the meantime we've found out about `404.html` that's generated from `404.js`. Spring Boot picks this one and serve it so we should probably do something for other errors.

Issue #185, opened 2019-07-08
Provide a link to additional starters in Help menu


mbhave:
The list of starters/dependencies available for project generation on start.spring.io is curated based on a number of factors. There are additional community starters that we could help discover by pointing to [this](https://github.com/spring-projects/spring-boot/blob/b0626bca320c21928e9790fcd2c59337344f9ac5/spring-boot-project/spring-boot-starters/README.adoc).

snicoll:
Closing in favour of PR #261

snicoll:
We need to rediscuss this one as the link between Spring Boot generations and 3rd party starters is not obvious. @oodamien was considering a different UI change so I've closed the PR

MDF-1951:
After reviewing Issue #185 and PR #261, I noticed the concern that the relationship between the selected Spring Boot version and the linked community starters was not obvious. Would maintainers be interested in revisiting this issue with a version-aware UI approach (for example, a dedicated page that explains the distinction between curated and community-maintained starters and can take the selected Spring Boot version into account) before I investigate a potential implementation?

mhalbritter:
We talked about it and decided that we no longer need this.

PR #261, opened 2019-09-10
Provide a link to additional starters in Help menu

oodamien:
See gh-185

### DIFF
--- a/start-client/src/components/common/layout/QuickLinks.js
+++ b/start-client/src/components/common/layout/QuickLinks.js
@@ -119,6 +119,21 @@ class QuickLinks extends React.Component {
                       What's New With Spring
                     </a>
                   </li>
+
+                  <li>
+                    <a
+                      id='ql-help-comminty'
+                      target='_blank'
+                      rel='noopener noreferrer'
+                      tabIndex='-1'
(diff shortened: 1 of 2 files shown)


start-client/src/components/common/layout/QuickLinks.js
```diff
@@ -119,6 +119,21 @@ class QuickLinks extends React.Component {
                       What's New With Spring
                     </a>
                   </li>
+
+                  <li>
+                    <a
+                      id='ql-help-comminty'
```
snicoll: Typo?

start-client/src/components/common/layout/QuickLinks.js
```diff
@@ -119,6 +119,21 @@ class QuickLinks extends React.Component {
...
+                      target='_blank'
+                      rel='noopener noreferrer'
+                      tabIndex='-1'
+                      href='https://github.com/spring-projects/spring-boot/blob/b0626bca320c21928e9790fcd2c59337344f9ac5/spring-boot-project/spring-boot-starters/README.adoc'
```
snicoll: This page is actually different depending on the Spring Boot version you are using. The link between this link and the Spring Boot selector is nonexistent so there's not much we can do about it.

That said, linking to a particular revision is not what I would do. I'd like to get more feedback.

snicoll:
As discussed on the call, we're considering a different approach that needs to be further discussed on the issue that I've reopened, see https://github.com/spring-io/start.spring.io/issues/185#issuecomment-538282797

PR #267, opened 2019-09-13
Revisit error message when client fails to connect

oodamien:
image

### DIFF
--- a/start-client/src/pages/index.js
+++ b/start-client/src/pages/index.js
@@ -204,7 +204,7 @@ class IndexPage extends React.Component {
       meta,
       deps
     ).catch(err => {
-      toast.error('The server API is not available.')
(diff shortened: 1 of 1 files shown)


start-client/src/pages/index.js
```diff
@@ -30,6 +30,14 @@ import {
 } from '../components/utils/api'
 import { getValidDependencies } from '../components/utils/versions'
 
+const ErrorToastMessage = () => (
+  <span>
+    Could not connect to <span className='link-website'>start.spring.io</span>.
```
snicoll: Sorry Damien, I don't think that's what we should do here. I was trying this locally and tried to get the popup and realized having "start.spring.io" while I am running on localhost looks super odd.

I was about to polish that myself but I can see you've added some styling so I want to make sure this is done properly.

How about "Could not connect to server. Please check your network".

PR #271, opened 2019-09-13
Change protocol of MyBatis reference page to https

kazuki43zoo:
We supported `https` on `mybatis.org`.


snicoll:
Thank you @kazuki43zoo

PR #272, opened 2019-09-16
Upgrade to Vaadin 14.0.4

tulioag:


snicoll:
@tulioag I just saw the notification and upgraded in https://github.com/spring-io/start.spring.io/commit/edb69f8816614f236307f03d4dd2185dd047b232. Thanks for the PR in any case.

Issue #278, opened 2019-09-21
Building maven project for spring cloud contract with kotlin


eddumelendez:
working in a project with `kotlin` and `spring-cloud-contract` the following changes were needed in a `maven` project. Contracts were defined in `kotlin` too

* Producer's `pom.xml`

```xml
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-contract-spec-kotlin</artifactId>
    <scope>test</scope>
</dependency>
<dependency>
    <groupId>io.rest-assured</groupId>
    <artifactId>spring-web-test-client</artifactId>
    <scope>test</scope>
</dependency>
```

Needed `io.rest-assured:spring-web-test-client` because was using `webflux`

```xml
<plugin>
    <groupId>org.apache.maven.plugins</groupId>
    <artifactId>maven-compiler-plugin</artifactId>
    <executions>
        <!-- Replacing default-compile as it is treated specially by maven -->
        <execution>
            <id>default-compile</id>
            <phase>none</phase>
        </execution>
        <!-- Replacing default-testCompile as it is treated specially by maven -->
        <execution>
            <id>default-testCompile</id>
            <phase>none</phase>
        </execution>
        <execution>
            <id>java-compile</id>
            <phase>compile</phase>
            <goals>
                <goal>compile</goal>
            </goals>
        </execution>
        <execution>
            <id>java-test-compile</id>
            <phase>test-compile</phase>
            <goals>
                <goal>testCompile</goal>
            </goals>
        </execution>
    </executions>
</plugin>
```

This dependency was also needed in `org.springframework.cloud:spring-cloud-contract-maven-plugin`

```xml
<dependency>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-contract-spec-kotlin</artifactId>
    <version>...</version>
</dependency>
```

Also, upgrade dependency `<rest-assured.version>4.0.0</rest-assured.version>` was required.

* Consumer's `pom.xml`

```xml
<dependency>
    <groupId>org.jetbrains.kotlin</groupId>
    <artifactId>kotlin-scripting-compiler-embeddable</artifactId>
    <version>${kotlin.version}</version>
</dependency>
```

I am wondering if this is something will be provided by `start.spring.io` or it will be documented in `HELP.md`

snicoll:
Thanks for the report @eddumelendez. We support tuning the build for such dependency so I think we should do our best to support that out of the box.

[...]

PR #283, opened 2019-09-27
Improvements regarding spring cloud contract and kotlin projects

eddumelendez:
See #278

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractMavenBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractMavenBuildCustomizer.java
@@ -16,7 +16,10 @@
 
 package io.spring.start.site.extension.dependency.springcloud;
 
+import io.spring.initializr.generator.buildsystem.Dependency;
 import io.spring.initializr.generator.buildsystem.maven.MavenBuild;
+import io.spring.initializr.generator.language.Language;
+import io.spring.initializr.generator.language.kotlin.KotlinLanguage;
 import io.spring.initializr.generator.project.ProjectDescription;
 import io.spring.initializr.generator.spring.build.BuildCustomizer;
 import io.spring.initializr.generator.version.Version;
@@ -28,10 +31,13 @@ import org.apache.commons.logging.LogFactory;
  * with Maven.
  *
  * @author Olga Maciaszek-Sharma
+ * @author Eddú Meléndez
  */
 class SpringCloudContractMavenBuildCustomizer implements BuildCustomizer<MavenBuild> {
 
 	private static final Log logger = LogFactory.getLog(SpringCloudContractMavenBuildCustomizer.class);
+
+	private static final Version VERSION_2_2_0_M2 = Version.parse("2.2.0.M2");
 
 	private final ProjectDescription description;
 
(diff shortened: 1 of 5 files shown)


snicoll:
Thanks for the PR @eddumelendez! I've asked the Spring Cloud Team to have a look to this because I don't understand why the lifecycle has to be changed like this. I wonder if there's a way to avoid that.

start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractMavenBuildCustomizer.java
```diff
@@ -57,7 +63,37 @@ public void customize(MavenBuild mavenBuild) {
 		mavenBuild.plugins().add("org.springframework.cloud", "spring-cloud-contract-maven-plugin", (plugin) -> {
 			plugin.version(sccPluginVersion);
 			plugin.extensions();
+
+			if (isKotlin(this.description.getLanguage()) && isSpringCloudVersionAtLeastAfter()) {
```
marcingrzejszczak: That makes sense cause you need to add a dependency

start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractMavenBuildCustomizer.java
```diff
@@ -57,7 +63,37 @@ public void customize(MavenBuild mavenBuild) {
...
+				plugin.dependency("org.springframework.cloud", "spring-cloud-contract-spec-kotlin", sccPluginVersion);
+			}
 		});
+
+		if (isKotlin(this.description.getLanguage()) && isSpringCloudVersionAtLeastAfter()) {
+			mavenBuild.dependencies().add("spring-cloud-contract-spec-kotlin",
+					Dependency.withCoordinates("org.springframework.cloud", "spring-cloud-contract-spec-kotlin"));
+
+			mavenBuild.plugins().add("org.apache.maven.plugins", "maven-compiler-plugin", (plugin) -> {
```
marcingrzejszczak: Why is it necessary? Is it because of Kotlin?

start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractStubRunnerMavenBuildCustomizer.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private final ProjectDescription description;
+
+	SpringCloudContractStubRunnerMavenBuildCustomizer(ProjectDescription description) {
+		this.description = description;
+	}
+
+	@Override
+	public void customize(MavenBuild mavenBuild) {
+		if (isKotlin(this.description.getLanguage()) && isSpringCloudVersionAtLeastAfter()) {
```
marcingrzejszczak: Hmm... it's required for the plugin dependency or in general? Cause maybe `spring-cloud-kotlin-spec` should provide this dependency if it doesn't.

snicoll:
So adding spring cloud contract triggers cross-compilation as the contract are generated in Java if I understood correctly. I am not sure what to do about that and this doesn't feel very idiomatic. Would it be possible to fix that in Spring Cloud Contract rather than having to enable all that machinery when a single dependency is selected?

start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractMavenBuildCustomizer.java
eddumelendez: without it, `ContractVerifierTest.java` can not resolve `BaseContractTest.kt`

I copy this configuration from an example at [spring-cloud-contract-samples](https://github.com/spring-cloud-samples/spring-cloud-contract-samples/blob/master/producer_kotlin/pom.xml#L107-L137)

start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractStubRunnerMavenBuildCustomizer.java
eddumelendez: I am not sure about having it at `spring-cloud-kotlin-spec` since it just fails for the consumer and not the producer. This is the message thrown when try to run the consumer without it.

```
java.lang.IllegalStateException: Failed to load ApplicationContext
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'batchStubRunner' defined in class path resource [org/springframework/cloud/contract/stubrunner/spring/StubRunnerConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.cloud.contract.stubrunner.BatchStubRunner]: Factory method 'batchStubRunner' threw exception; nested exception is java.lang.NoClassDefFoundError: org/jetbrains/kotlin/scripting/compiler/plugin/repl/GenericReplCompiler
Caused by: org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.cloud.contract.stubrunner.BatchStubRunner]: Factory method 'batchStubRunner' threw exception; nested exception is java.lang.NoClassDefFoundError: org/jetbrains/kotlin/scripting/compiler/plugin/repl/GenericReplCompiler
Caused by: java.lang.NoClassDefFoundError: org/jetbrains/kotlin/scripting/compiler/plugin/repl/GenericReplCompiler
```

snicoll:
We've discussed this and we're not keen to bring such complexity to the build. The typical example I have in mind is going down the rabbit hole with another dependency requiring such involved tweak to the build setup, potentially incompatible with each other.

We believe this is something that should be fixed in Spring Cloud Contract, I've opened https://github.com/spring-cloud/spring-cloud-contract/issues/1243

Thanks for the PR @eddumelendez and sorry for the wasted effort.

eddumelendez:
don't worry @snicoll at least it was useful for the discussion :)

I will follow the issue you created

PR #310, opened 2019-10-21
Add support for Spring Cloud LoadBalancer

OlgaMaciaszek:


start-site/src/main/resources/application.yml
```diff
@@ -1028,6 +1028,16 @@ initializr:
           description: Declarative REST Client. OpenFeign creates a dynamic implementation of an interface decorated with JAX-RS or Spring MVC annotations.
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-openfeign
+        - name: Spring Cloud LoadBalancer
+          id: cloud-loadbalancer
+          compatibilityRange: "[2.2.0.M4,2.2.x.BUILD-SNAPSHOT)"
```
snicoll: Is there a reason why the entry should be disabled with snapshots? (Compared to the rest of Spring Cloud I mean).

start-site/src/main/resources/application.yml
```diff
@@ -1028,6 +1028,16 @@ initializr:
...
+          description: Client-side load-balancing with Spring Cloud LoadBalancer.
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-loadbalancer
+          links:
+            - rel: guide
```
snicoll: Can we please have a link to the documentation as well?

start-site/src/main/resources/application.yml
OlgaMaciaszek: No, for 2.2.x snapshots it should be enabled - if that's the result of it, then it's a bug; should this look differently to include 2.2.x snapshots?

snicoll: That code says the entry should be enabled beteen `2.2.0.M4` (inclusive) to the latest snapshot of the 2.2 line (excluded). We don't use such range but rather the version one should use at the bare minimum.

start-site/src/main/resources/application.yml
OlgaMaciaszek: It's not a separate project, so documentation for it is included in the SC Commons docs - there are different bits related to using it in different scenarios:
- https://cloud.spring.io/spring-cloud-static/spring-cloud-commons/2.2.0.M3/reference/html/#spring-resttemplate-as-a-load-balancer-client
- https://cloud.spring.io/spring-cloud-static/spring-cloud-commons/2.2.0.M3/reference/html/#spring-resttemplate-as-a-load-balancer-client
- https://cloud.spring.io/spring-cloud-static/spring-cloud-commons/2.2.0.M3/reference/html/#passing-your-own-load-balancer-client-configuration
Should I just add all of them?

snicoll: Can we rephrase the doc a bit to have a dedicated section for it please? If there is a dedicated starter, that's what I would expect.

start-site/src/main/resources/application.yml
OlgaMaciaszek: ah, ok. thanks

start-site/src/main/resources/application.yml
OlgaMaciaszek: Have merged the changes: https://github.com/spring-cloud/spring-cloud-commons/pull/624 Once we do a new release (should be this week) and this gets published, will add it here.

Issue #297, opened 2019-10-17
Provide the right testMode when webflux and spring cloud contract are selected


eddumelendez:
`spring-cloud-contract-maven-plugin` provides a way to set the `testMode`. if `webflux` is selected the `testMode` must be `WEBTESTCLIENT`

```xml
 <plugin>
    <groupId>org.springframework.cloud</groupId>
    <artifactId>spring-cloud-contract-maven-plugin</artifactId>
    <version>2.2.0.M2</version>
    <extensions>true</extensions>
    <configuration>
        <testMode>WEBTESTCLIENT</testMode>
    </configuration>
</plugin>
```

snicoll:
Closing in favour of PR #313

PR #313, opened 2019-10-24
Add testMode property when webflux and scc verifier are selected

eddumelendez:
See gh-297


snicoll:
Thanks again @eddumelendez

PR #327, opened 2019-11-10
Add reactive facet to RSocket entry

wonwoo:
I think reactor-test should added when select rsocket.

What do you think?


snicoll:
Seems fair, thanks a lot for the PR!

Issue #303, opened 2019-10-18
Remove Gatsby


oodamien:
Implement a React App, simplify the frontend techno stack.

snicoll:
Closing in favour of PR #330

snicoll:
This lead to a regression that the tests didn't catch (see #337). I'd like us to rework this in such a way that a test would have broken.

snicoll:
This is now superseded by #352

PR #330, opened 2019-11-13
Remove Gatsby

oodamien:
New optimized build
Remove Gatsby
Rewrite with React Hook
Autocomplete with Fuse.js
see gh-303


snicoll:
I don't think the new searcher is ready just yet. I get no result for `stream`. Can we please get the build refactoring without the new search feature?

snicoll:
This lead to regression so this PR has been reverted

PR #345, opened 2019-12-04
Upgrade to Vaadin 14.1.3

ZheSun88:


ZheSun88:
This is a minor update do we need to change the `compatibilityRange`? and where I can get that info?

mbhave:
@ZheSun88 I don't think we'd want to replace `14.0.15` which works with Spring Boot `2.1.x`, would we? I think we'd want another entry for `14.1` which has a compatibility range signifying that it works with Spring Boot `2.2.x`.

snicoll:
I agree that we should add a new entry for this release. @ZheSun88 if you were asking how to do that, please check [the documentation](https://docs.spring.io/initializr/docs/current-SNAPSHOT/reference/html/#howto-link-boot-version). If you agree, we can take care of that when merging the PR.

(We can't at the moment, see https://github.com/spring-io/start.spring.io/issues/348).

ZheSun88:
Hi @snicoll and @mbhave ,

yes, I think it would be good to have a new entry for the new minor. the only question is whether we should keep `14.0.15` there or not, as it is not going to be supported anymore. 

@mstahv , what do you think?

snicoll:
If we remove `14.0.15` then it's the same as replacing the current mapping. The question is whether we should go 14.1 for Spring Boot 2.2+ and 14.0 for Spring boot 2.1.+. If the new feature release immediately supersedes 14.0 then we should simply replace it. Maybe a link to your support policy would help drive the discussion?

ZheSun88:
the current release train mode is `14.0` will be replaced after 14.1 got a stable release. 

this is what we mentioned in the release note
```
This is a stable release of Vaadin 14.1 that replaces Vaadin 14.0 in the 14.X LTS series.
```

snicoll:
Well that settle it then. 14.1 should replace 14.0.

ZheSun88:
so can I leave this for the team to handle the update?

snicoll:
I didn't get that. Which team? If you're talking about your PR, it already replaces 14.0 with 14.1

ZheSun88:
```
If you agree, we can take care of that when merging the PR.
```
i meant this one.

snicoll:
That was about keeping 14.0 and 14.1 together. We just decided not to do that so your PR is doing the right thing already. Have I missed something?

ZheSun88:
I am asking that because I didn't update the `compatibilityRange: "2.1.0.M1"`

snicoll:
You lost me again. If 14.1 replaces 14.0 then no replacement of the compatibility range is required. To avoid more back and forth can you please state what the problem with your current PR is?

ZheSun88:
``` 
If 14.1 replaces 14.0 then no replacement of the compatibility range is required.
```
This should be the answer I need. 

My question was, should we update the compatibility range, because the `spring-boot` version used in `14.1` is not `2.1.0` but `2.2.0.RELEASE` https://github.com/vaadin/spring/commit/1bb45f9b4ec185588e1ba6f3c526bdbe53739982

for me, it looks the `compatibility Range` is misleading me.

snicoll:
>  the spring-boot version used in 14.1 is not 2.1.0 but 2.2.0.RELEASE vaadin/spring@1bb45f9

That makes it an obvious candidate for 2.2+ then. The `compatiblityRange` defines the version range of Spring Boot that this version of Vaadin should use. 

Is `14.0` EOL or are you still providing support for it? I really don't know what we can do here. If 14.1 is built against SB 2.2, then we should not offer it for 2.1. If 14.0 is EOL then we shouldn't offer it here. Our only option is to disable Vaadin with SB < 2.2 and I really would like to avoid that.

I'd argue it's also a problem for your customers right now if they don't have a Vaadin version that's officially supported against Spring Boot 2.1.

ZheSun88:
I will contact my team about this change, but just answer the question based on what i have known:

yes, 14.0 is EOL. Vaadin 14 is long term support version, but it will be divided into `14.0`, `14.1` and later `14.2` and so on.

ZheSun88:
Asked the team, we can keep it in this way without change compatibilityRange.

snicoll:
Sorry but that doesn’t work for us. If that version builds against 2.2, do you have a CI job that builds it against 2.1 as well? If you don’t, I am not keen to use that version with 2.1. 

Some clarification on the upgrade policy would be most welcome as well. Thanks!

pleku:
Hi. Sorry for the confusing messaging from our side, but it quite clear:
Vaadin 14.1.x works with **both** Spring Boot 2.1 & 2.2. Both are validated.

We are building things on top of Spring Boot 2.2 since we want to stay on the cutting edge and discover issues early.

snicoll:
Thanks for the feedback @pleku! That’s great news, thank you all for your efforts on supporting the latest spring releases.

PR #350, opened 2019-12-07
Polish Apache Camel description

davsclaus:


start-site/src/main/resources/application.yml
```diff
@@ -718,7 +718,7 @@ initializr:
               groupId: org.apache.camel
             - compatibilityRange: "2.2.0.M1"
               version: 3.0.0
-          description: Apache Camel lets you create the Enterprise Integration Patterns to implement routing and mediation rules a Java based Domain Specific Language via Spring.
+          description: Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.
```
snicoll: We'd like to refer to the project as "Apache Camel".  Is there a reason you've removed the Apache trademark here?

davsclaus: Oh no reason it was a copy from the website, yeah lets add Apache that is better, thanks.

davsclaus: Did a commit with the update

Issue #365, opened 2020-01-10
Add reactor-kotlin-extensions to dependencies when creating a Kotlin + Webflux application


hantsy:
I was trying to upgrade [my sample](https://github.com/hantsy/spring-kotlin-coroutines-sample) to the following newest stack:

* Spring 2.2.2.RELEASE
* Java 11
* Kotlin/Webflux

When I tried to refresh the APIs according to the changes in the new dependencies and  found 

`WebClient....exchange().test()` is marked as **deprecated**, and the new *reactor.kotlin.test* is from `reactor-kotlin-extensions`.

I think it should be added into dependencies when creating a Kotlin based Webflux app, because all Kotlin extensions for Reactor is moved to this archetype.

```xml
<!-- https://mvnrepository.com/artifact/io.projectreactor.kotlin/reactor-kotlin-extensions -->
<dependency>
          <groupId>io.projectreactor.kotlin</groupId>
          <artifactId>reactor-kotlin-extensions</artifactId>
          <version>1.0.1.RELEASE</version>
</dependency>
```

snicoll:
Closing in favour of PR #391

PR #391, opened 2020-02-20
Add reactor-kotlin-extensions dependency

eddumelendez:
When the project has a reactive facet and the language is kotlin the
dependency `reactor-kotlin-entensions` will be added.

See gh-365


snicoll:
Thanks @eddumelendez. I've polished things a bit in 1c413f5

PR #416, opened 2020-03-28
RestAssured's WebClient support is not added although Spring Cloud Contract's testMode is set to use it

eddumelendez:


eddumelendez:
@snicoll what do you think about this PR? Your input would be appreciated.

snicoll:
@eddumelendez I apologize for the delay, I'll have a look to it shortly.

Paging @marcingrzejszczak

marcingrzejszczak:
Makes perfect sense! LGTM

snicoll:
Spring Cloud supports three modes as far as I understood:

* Spring MVC with restassured's mockmvc support
* Spring Webflux with restassured's web client support
* JAXRS with jaxrs support

As far as I understand, this PR provides support for only one of the three options which would make the support inconsistent. @eddumelendez have I missed something here?

eddumelendez:
> * Spring MVC with restassured's mockmvc support

Currently, when a project is generated, by default this is supported.

> * Spring Webflux with restassured's web client support

This PR adds support to add the missing dependencies to work in a project with webflux

The last one is missing.

snicoll:
Well you could argue that if you had `webtestclient` support, maybe you should exclude the MockMvc one then? I have a hard time resonate on the workflow here.

snicoll:
On the other hand, I can see that we set the `testMode` flag when webflux is present so not adding the dependency is really odd. I'll do that but I wish we can revisit the support at some point @marcingrzejszczak

snicoll:
Thanks for the PR @eddumelendez, I've polished it to make the dependency more explicit with the fact we're switching the test mode (see https://github.com/spring-io/start.spring.io/commit/b8073c2044921cf407bf669a161f3d96116fa92f). It's particularly important as adding the way you did would break users creating a project with Spring Boot 2.1.x as we don't have dependency management for rest assured there.

eddumelendez:
@snicoll thanks! Please, count with me in order to revisit the support. I will be glad to work on it.

PR #452, opened 2020-05-29
Do not call FileSaver if project generation failed

### LINKED ISSUE #451: If "Generate" hits an error the FileSaver is still called
If there's an error on "Generate" you get the expected popup error message, but you also get an ugly log in the console:

```
FileSaver.min.js:1 Uncaught (in promise) TypeError: Failed to execute 'createObjectURL' on 'URL': No function was found that matched the signature provided.
    at Function.a [as saveAs] (FileSaver.min.js:1)
    at Application.js:77
...
```

I think that's probably because the `onSubmit` catches the error, processes it, but doesn't prevent the call to `FileSaver`.

dsyer:
Fixes #451.


start-client/src/components/Application.js
```diff
@@ -71,10 +71,15 @@ export default function Application() {
       get(dependencies, 'list')
     ).catch(() => {
       toast.error(`Could not connect to server. Please check your network.`)
-      setGenerating(false)
     })
     setGenerating(false)
-    FileSaver.saveAs(project, `${get(values, 'meta.artifact')}.zip`)
+    if (project) {
+      if (project.type == "text/plain") {
+        project.text().then(text => toast.info(text))
```
oodamien, line 78: There is already an error message (generic), this info message is useless.

dsyer: Currently, yes. But it provides a useful option if the backend wanted to send a more specific message. It's not hurting anyone, and it might help.

oodamien: `getProject` return a promise (without data/message in case of rejection).
`project.type` will never be true.

dsyer: Sure. At least not with the current implementation of the backend. I was just suggesting an enhancement. Feel free to take it out if you feel strongly about it.

oodamien: Thanks Dave. I pushed a fix on the master (no ZIP creation in case of error).
https://github.com/spring-io/start.spring.io/commit/c425ac3a3ea9607c5bf94933576e0bf67e4badbb

oodamien:
Thanks Dave.
I pushed a fix on the master in case of service error. (https://github.com/spring-io/start.spring.io/commit/c425ac3a3ea9607c5bf94933576e0bf67e4badbb)

PR #467, opened 2020-06-11
Upgrade to Vaadin 14.2.1

manolo:


snicoll:
Thanks for the follow-up @manolo

PR #470, opened 2020-06-11
Add Spring Cloud App Broker

royclarkson:


start-site/src/main/resources/application.yml
```diff
@@ -1020,6 +1020,21 @@ initializr:
             - rel: guide
               href: https://github.com/spring-cloud-samples/bookstore-service-broker
               description: Using Spring Cloud Open Service Broker
+        - name: App Broker
```
snicoll, line 1023: While most Spring Cloud entries are unqualified, I find `App Broker` hard to resonate about, I don't know if `Cloud App Broker` would help with that.

start-site/src/main/resources/application.yml
```diff
@@ -1020,6 +1020,21 @@ initializr:
...
+          id: app-broker
```
snicoll, line 1024: This sounds like a very generic ID to me. Given that it's open api specific, it should probably be qualified.

start-site/src/main/resources/application.yml
```diff
@@ -1020,6 +1020,21 @@ initializr:
...
+          compatibilityRange: "[2.2.0.RELEASE,2.4.0-M1)"
+          description: Framework for building Spring Boot applications that implement the Open Service Broker API and deploy applications as brokered services.
+          groupId: org.springframework.cloud
+          artifactId: spring-cloud-starter-app-broker-cloudfoundry
```
snicoll, line 1028: Looking at the current description and name, it is not obvious at all this is Cloudfoundy specific (as I am guessing given the name of the starter).

start-site/src/main/resources/application.yml
royclarkson: `Cloud App Broker` works for me if you feel like it makes it more clear. Spring Cloud App Broker is an opinionated implementation framework of Spring Cloud Open Service Broker (or the Open Service Broker API).

start-site/src/main/resources/application.yml
royclarkson: is `osbapi-app-broker` better?

start-site/src/main/resources/application.yml
royclarkson: App Broker offers an abstraction, but it currently only supports Cloud Foundry. I'll add a bit of clarification.

snicoll:
Thanks Roy. I had mentioned this in the review

> I've looked at the suggestion and documentation and I am a bit confused about the use case and the developer experience of starting a new project including this dependency. Can you please provide some more info?

Can you please provide us with that information?

royclarkson:
@snicoll I'm closing this for now, but may revisit again in the future. Thanks for the feedback.

PR #483, opened 2020-07-10
Add Neo4j Bolt driver option.

meistermeier:
As a completion (or sidekick) for the changes in https://github.com/spring-projects/spring-boot/pull/22299 we would purpose to bring direct access to the Neo4j Java driver into the initializer.
This aligns with the already existing PostgreSQL, Oracle, etc. support and is a hugely demanded feature we hear from our community and customers. The background is that those users do not want to have to opt-in for a Spring Data (Neo4j) to just get a managed driver bean.

Looking forward to your feedback.

(ps: I love the easiness of adding configuration options in the initializer project 👍 )

### DIFF
--- a/start-client/dev/api.json
+++ b/start-client/dev/api.json
@@ -646,6 +646,11 @@
                 "templated": true
               }
             }
+          },
+          {
+            "id": "neo4j",
+            "name": "Neo4j Bolt driver",
+            "description": "A Neo4j Java driver that connects to a Neo4j instance or cluster using the Bolt protocol."
           }
         ]
       },
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -666,6 +666,12 @@ initializr:
               description: Accessing Data with Neo4j
             - rel: reference
               href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#boot-features-neo4j
+        - name: Neo4j Bolt driver
+          id: neo4j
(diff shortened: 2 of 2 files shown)


wilkinsona:
Thanks for your patience.

We discussed this one today and concluded that, while we acknowledge that the bolt driver can be used as a JDBC driver, we're concerned that its addition as a standalone option will be confusing. We think that people will wonder why it's offered but other stores like Cassandra and MongoDB are only offered via their Spring Data support. There's also no support for the JDBC side of things in Spring Boot so if someone added it expecting to use the Bolt driver's JDBC support they would not get the outcome they had anticipated.

Thanks anyway for the proposal and once again for your patience.

meistermeier:
Thanks for the feedback.
Just FYI: _Bolt driver's JDBC_ does not exist. To be more precise: There is no reference to jdbc in the driver at all.
FWIW why not adding e.g. Mongo, Cassandra, etc. configuration options? The versions are already managed in Spring Boot and I think there is a significant amount of users just needing drivers and not Spring Data <module>.

wilkinsona:
Sorry, I suspect I was mixing things up with https://github.com/neo4j-contrib/neo4j-jdbc.

There's nothing stopping someone from writing a Boot application form depending on the Cassandra, Neo4j, etc drivers without using Spring Data. We chose not to offer them here as we believe those users are in a minority and offering both options to everyone would require everyone to bear the cost of considering both and understanding the difference between the two. Our opinion is that the cost is too high for the benefit that it'll bring.

PR #509, opened 2020-08-10
Upgrade to Spring Boot Admin 2.3.0

joshiste:


start-site/src/main/resources/application.yml
```diff
@@ -47,8 +47,10 @@ initializr:
             version: 2.0.6
           - compatibilityRange: "[2.1.0.M1,2.2.0.M1)"
             version: 2.1.6
-          - compatibilityRange: "2.2.0.M1"
+          - compatibilityRange: "[2.2.0.M1,2.3.0.M1)"
             version: 2.2.4
+          - compatibilityRange: "2.3.0.M1"
+            version: 2.3.0
```
snicoll, line 53: Is SBA compatible/tested against Spring Boot 2.4 yet? It's probably a bit early but an open-ended range means they could use that combination.

joshiste: hmm sorry didn't have the time to test so.

snicoll: No worries, I'll polish the range.

PR #529, opened 2020-09-04
Fixing issue #522 - instance of ProjectGenerationInvoker is created per test method

### LINKED ISSUE #522: ConcurrentModificationException in integration tests
CI builds fail from time to time with 

```
java.util.ConcurrentModificationException
	at io.spring.start.site.ProjectGenerationIntegrationTests.projectBuilds(ProjectGenerationIntegrationTests.java:170)
```

It happens more often on my box.

mkrzywanski:
This pull request fixes #522 . The instance of `ProjectGenerationInvoker` is created per method invocation to get rid of `ConcurrentModificationException` because `ProjectGenerationInvoker` is not thread safe as it uses `LinkedHashMap` internally without any synchornization.

### DIFF
--- a/start-site/src/test/java/io/spring/start/site/ProjectGenerationIntegrationTests.java
+++ b/start-site/src/test/java/io/spring/start/site/ProjectGenerationIntegrationTests.java
@@ -99,14 +99,13 @@ class ProjectGenerationIntegrationTests {
 		}
 	});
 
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR but I’d rather fix the tread safety. What do you think?

mkrzywanski:
Sure, will apply the fix in the other project then.

snicoll:
Perfect thanks!

PR #530, opened 2020-09-06
Upgrade to Apache Camel 3.5.0

davsclaus:
Camel 3.5 was released: https://camel.apache.org/blog/release-3-5-0.html

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -782,7 +782,7 @@ initializr:
             - compatibilityRange: "[2.2.0.M1,2.3.0.M1)"
               version: 3.3.0
             - compatibilityRange: "[2.3.0.M1,2.4.0-M1)"
-              version: 3.4.3
+              version: 3.5.0
           description: Apache Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.
           groupId: org.apache.camel.springboot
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -782,7 +782,7 @@ initializr:
             - compatibilityRange: "[2.2.0.M1,2.3.0.M1)"
               version: 3.3.0
             - compatibilityRange: "[2.3.0.M1,2.4.0-M1)"
-              version: 3.4.3
+              version: 3.5.0
```
snicoll, line 785: `2.3.x` is in maintenance phase and Spring Boot 2.4 will be released in a month or so. Would it be worth mapping this version with Spring Boot 2.4 rather than bumping to a new feature release now?

snicoll (commented):
Thanks for the PR. Wondering what your plan are for Spring Boot 2.4, I've added a comment.

start-site/src/main/resources/application.yml
davsclaus: Camel 3.6 will upgrade to SB 2.4 and is planned for 6-8 weeks or so.

snicoll: Great news, thanks for the feedback @davsclaus

snicoll (approved):

PR #532, opened 2020-09-15
Upgrade to Vaadin 14.3.6

ZheSun88:


snicoll:
Thanks again for the follow-up @ZheSun88, much appreciated.

PR #575, opened 2020-11-20
Update compatibility range for 'okta'

bdemers:
The Okta Spring Boot Starter v1.5.0 has a minimum version compatibility with Boot 2.4

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -429,14 +429,12 @@ initializr:
           description: Okta specific configuration for Spring Security/Spring Boot OAuth2 features. Enable your Spring Boot application to work with Okta via OAuth 2.0/OIDC.
           groupId: com.okta.spring
           artifactId: okta-spring-boot-starter
-          compatibilityRange: "[2.1.2.RELEASE,2.4.0-M1)"
+          compatibilityRange: "2.4.0"
           mappings:
             - compatibilityRange: "[2.1.2.RELEASE,2.2.0.M1)"
               version: 1.2.1
             - compatibilityRange: "[2.2.0.M1,2.4.0-M1)"
               version: 1.4.0
-            - compatibilityRange: "[2.4.0,)"
-              version: 1.5.0
           links:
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -429,7 +429,7 @@ initializr:
           description: Okta specific configuration for Spring Security/Spring Boot OAuth2 features. Enable your Spring Boot application to work with Okta via OAuth 2.0/OIDC.
           groupId: com.okta.spring
           artifactId: okta-spring-boot-starter
-          compatibilityRange: "[2.1.2.RELEASE,2.4.0-M1)"
+          compatibilityRange: "2.4.0"
```
bdemers, line 432: @snicoll is this the correct format to set the minimum version, or _should_ it be something like `[2.4.0,)`?

snicoll: I am not keen to have open ranges for 3rd party libraries so I'll polish that to be 2.4.x so that the entry is disabled with `2.5.x`. As always if you have a CI build that works with that generation just ping us and we'll relax the range.

snicoll:
This PR misses the mapping for version 1.5.0 or am I missing something?

start-site/src/main/resources/application.yml
snicoll: That said You've just removed support for 2.2.x and 2.3.x. Why do you want to remove the minimum version to be 2.4.x as we have previous okta versions that work fine with those Spring Boot generations?

Can you please review [the documentation](https://docs.spring.io/initializr/docs/current/reference/html/#dependencies-compatibility-range)?

snicoll:
Thanks for the PR but the proposed changes are not adding 1.5.0 and are breaking use cases for 2.2.x and 2.3.x. Let's follow-up on #576.

start-site/src/main/resources/application.yml
bdemers: Super helpful! thank you!

PR #577, opened 2020-11-23
Adds Spring Cloud GCP 2.0 module and BOM

ttomsu:
With the pending release of Spring Cloud Ilford, this PR adds the Spring Cloud GCP 2.0 module (still in release candidate form). 

Opening this as a draft PR for now with RC and SNAPSHOTs, but will remove those when RC2 is released with full Spring Boot 2.4 support.

cc/ @meltsufin @elefeint

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -157,6 +157,14 @@ initializr:
             version: 2.2.6.RELEASE
           - compatibilityRange: "[2.3.0.RELEASE,2.4.0-M1)"
             version: 2.3.0.RELEASE
+      spring-cloud-gcp:
+        groupId: com.google.cloud
+        artifactId: spring-cloud-gcp-dependencies
+        versionProperty: spring-cloud-gcp.version
+        additionalBoms: [spring-cloud]
+        mappings:
+          - compatibilityRange: "[2.4.0,2.4.1)"
+            version: 2.0.0-RC2-SNAPSHOT
       spring-geode:
         groupId: org.springframework.geode
         artifactId: spring-geode-bom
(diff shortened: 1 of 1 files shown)


snicoll:
@ttomsu there is a checklist to follow to require the additions of new dependencies so please create an issue and follow the instructions there.

start-site/src/main/resources/application.yml
```diff
@@ -1410,6 +1418,43 @@ initializr:
             - rel: guide
               href: https://github.com/spring-cloud/spring-cloud-gcp/tree/master/spring-cloud-gcp-samples/spring-cloud-gcp-storage-resource-sample
               description: GCP Storage Sample
+    - name: Google Cloud Platform 2.0
```
snicoll, line 1421: Note for later. I don't think we should have one entry for what's in the Spring Cloud release train and a 2.0 entry for whatever is next. Rather it must be the same entry and initializr would flip the coordinates based on the chosen Spring Boot generation. I am not sure that the library has all that it needs to do that so let's take a step back and, rather than discussing on a proposal, let's discuss the use  cases we're trying to achieve here. Some sample build snippet would be helpful so that we know what we have to generate.

ttomsu: I originally put them into the same module, but the `bom` field is 1.) only available at the top level (AFAICT) and 2.) only applicable to the 2.0 module.

snicoll: Thanks but that's why I prefer that we discuss on the use cases. I am not keen to duplicate the entries so we'll have to find another solution. Back to my original comment on not submitting a PRs for what I understood at first as new entries: are those a 1 to 1 mapping to what we had when GCP was part of the release train?

meltsufin: +1 on let's not duplicate entries but automatically detect based on version.

ttomsu: Issue created: https://github.com/spring-io/start.spring.io/issues/578

I'd also rather not have duplicate entries, but I was working against what was already built.

> are those a 1 to 1 mapping to what we had when GCP was part of the release train?

Yes, these are the 3 modules currently listed under GCP now - the "everything" module, pub/sub, and storage. We'd love to add more if you're willing to allow more.

snicoll:
Alright, I am going to close this now as we need to find a way to keep the same entries. Let's continue the discussion on the issue you've created.

PR #579, opened 2020-11-24
Upgrade to Vaadin 14.4.3

ZheSun88:


snicoll:
Thanks for the update @ZheSun88. Have you noticed that Vaadin is disabled with Spring Boot 2.4? It would be nice if you could get a version of `vaadin-spring` (and the related starter).

/cc @mstahv

pleku:
@snicoll thanks for the heads up! We just tested and the same `vaadin-spring` version used with Spring Boot 2.3 works just fine with 2.4.0. We've updated our builds to use 2.4.0.

snicoll:
@pleku thank you for the prompt feedback. We'll continue with this version then.

PR #601, opened 2020-12-27
Fix validation entry description to refer to Bean Validation API

ksundong:
The Validation Spec description not JSR-303 but JSR-380.
So I updated description JSR-303 to JSR-380.

I was confused about this description. 'JSR-303 validation with Hibernate validator.'
Spring Boot Starter Validation is using jakarta bean validation 2.0(JSR-380).
So, I think this description could be changed.

Thank you for this project!

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -859,7 +859,7 @@ initializr:
               href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#howto-batch-applications
         - name: Validation
           id: validation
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR.

> I was confused about this description. 'JSR-303 validation with Hibernate validator.'

That's also the bean validation API pre JakartaEE era but you're right it's a little bit outdated.

snicoll:
@ksundong thank you for making your first contribution to start.spring.io. I've used the opportunity to polish the entry further and remove any reference to a JSR, see https://github.com/spring-io/start.spring.io/commit/3f9ebb1116d21af10431b8836f190d869232caff

PR #603, opened 2020-12-28
Upgrade to Okta Spring Boot 2.0.0

arvindkrishnakumar-okta:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -446,7 +446,7 @@ initializr:
             - compatibilityRange: "[2.2.0.M1,2.4.0-M1)"
               version: 1.4.0
             - compatibilityRange: "[2.4.0-M1,2.5.0-M1)"
-              version: 1.5.1
+              version: 2.0.0
           links:
             - rel: guide
               href: https://github.com/okta/samples-java-spring/tree/master/okta-hosted-login
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -445,7 +445,9 @@ initializr:
               version: 1.2.1
             - compatibilityRange: "[2.2.0.M1,2.4.0-M1)"
               version: 1.4.0
-            - compatibilityRange: "[2.4.0-M1,2.5.0-M1)"
+            - compatibilityRange: "[2.4.0-M1,2.4.1)"
+              version: 1.5.1
+            - compatibilityRange: "[2.4.1,2.5.0-M1)"
               version: 2.0.0
           links:
             - rel: guide


pivotal-issuemaster:
@arvindkrishnakumar-okta Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-io/start.spring.io&pullRequestId=603)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-io/start.spring.io&pullRequestId=603) to manually synchronize the status of this Pull Request.

See the [FAQ](https://cla.pivotal.io/about) for frequently asked questions.

pivotal-issuemaster:
@arvindkrishnakumar-okta Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-io/start.spring.io&pullRequestId=603)!

snicoll:
@arvindkrishnakumar-okta 2.0.0 seems like a major rewrite to me. This would bring that to users who were previously on 1.5 and the default choice. Is that what you intended to do?

It's fine by us, just wanted to double check.

start-site/src/main/resources/application.yml
```diff
@@ -446,7 +446,7 @@ initializr:
             - compatibilityRange: "[2.2.0.M1,2.4.0-M1)"
               version: 1.4.0
             - compatibilityRange: "[2.4.0-M1,2.5.0-M1)"
-              version: 1.5.1
+              version: 2.0.0
```
snicoll, line 449: To be consistent with the [version matrix](https://github.com/okta/okta-spring-boot/blob/master/README.md#spring-boot-version-compatibility), I think this should be changed.

It's a bit of strange mapping, hence why I asked the question earlier but the proper metadata would be:

```yaml
- compatibilityRange: "[2.4.0-M1,2.4.1)"
  version: 1.5.1
- compatibilityRange: "[2.4.1,2.5.0-M1)"
  version: 2.0.0
```

We can't have overlaps or holes in the range so that looks more consistent to me. Thoughts?

arvindkrishnakumar-okta: @snicoll thanks! the one you propose above looks good.

arvindkrishnakumar-okta:
> @arvindkrishnakumar-okta 2.0.0 seems like a major rewrite to me. This would bring that to users who were previously on 1.5 and the default choice. Is that what you intended to do?
> 
> It's fine by us, just wanted to double check.

Updated the PR with your recommendation.

snicoll (approved):

snicoll:
Thanks for the follow-up and for making your first contribution to start.spring.io.

Issue #600, opened 2020-12-23
Improve end-user experience with Spring Integration


artembilan:
Currently Spring Integration appears in the search result only when we type something like "integration", "messaging", "enterprise", "adapter" - some general words in the project description or tags. I suggest to improve that text for all the support integration protocols: (s)ftp, JMS, JDBC, RSocket, ZeroMQ etc. You name it! See more about supported in the project modules: https://github.com/spring-projects/spring-integration. This way end-user would know that Spring Integration supports this or that technology according their wishes.

Another improvement suggestion is to add respective Spring Integration module as dependency when some target well-known library is supported in Spring Integration. For example when end-user asks for AMQP, we suggest Spring Integration (if the first point is fixed) and Spring for RabbitMQ. When end-user select both, we also add implicitly `spring-integration-amqp`, so we assume that in most cases users want to use channel adapters for AMQP from Spring Integration. And so on for all the supported protocols. Some similar solution I see here: https://github.com/spring-io/start.spring.io/blob/master/start-site/src/main/java/io/spring/start/site/extension/dependency/springsession/SpringSessionBuildCustomizer.java.

Another proposal, which might be too destructive, is a "fine-graining" Spring Integration modules, like you do at the moment for Spring Data. So, this is just like another angle looking at the original end-user experience problem...

If any thing of this sounds good, let me know and I'll issue a PR!

Happy holidays!

snicoll:
Thanks for the suggestion Artem. There are several things to unwind here and we prefer to focus on one thing at time so we may fork other issues based on the discussion.

IMO, the key to your proposal is to add a spring-integration specific module when a given tech is selected. This brings the most value and helps connect the dots between the technology and what Spring Integration has to offer. To move forward with this, could you please provide us with a mapping between an entry on start.spring.io and the `artifactId` of the spring-integration module to add? Said module should be in the bom which I guess is a given. If there is some "module-specific" section in the documentation, please add that to the mapping as we could augment `HELP.md` with that based on what the use has selected.

Based on this list, we can build that smart mapping and improve the description on start.spring.io. We need to be careful to avoid a too lengthy description for the Spring Integration entry so we may use another way.

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

[...]

PR #605, opened 2021-01-05
Include Spring Integration specific modules when a matching entry is selected

artembilan:
Fixes https://github.com/spring-io/start.spring.io/issues/600

Introduce an infrastructure for Spring Integration modules
and add respective dependencies into the generated POM
& doc links into a HELP.md when specific existing entry
is selected from the start.spring.io

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springintegration/SpringIntegrationBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springintegration/SpringIntegrationBuildCustomizer.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2012-2021 the original author or authors.
+ *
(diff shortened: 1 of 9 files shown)


artembilan:
New infrastructure classes are modeled after `testcontainers` modules. If that will simplify review somehow...
Thanks

snicoll:
Thanks @artembilan. Reviewing the PR, I found that we need an infrastructure that would reduce the amount of copy/paste so I did that in #607, rebased and use that rather than the copy. I've also updated the tests to use parameterised tests like the testcontainers were.

This should be available in production once the smoke tests pass. Thank you for your contribution!

artembilan:
Yes, Stephane, I thought about copy/paste problem, but then I decided to go with the final goal to make it easier for your to review.
I'm glad that you have figured out a common pattern and infrastructure. Hope it was not so hard for you 😉 

Thanks for the `@ParameterizedTest` trick! 😄

PR #296, opened 2019-10-17
Align spring cloud contract plugin configuration with JUnit generation

eddumelendez:
Spring Boot 2.2 provides JUnit 5 by default. In order to generate
tests with JUnit 5 the right setup is needed. This commit introduces
the setup for maven and gradle.

See gh-294

PR #613, opened 2021-01-11
Invalid Spring Cloud Contract Gradle plugin configuration with Spring Cloud 2020.0.x

### LINKED ISSUE #612: Spring Cloud Contract Gradle plugin configuration does not work with the latest version of the plugin
There are two issues with the generated configuration for Spring Cloud Contract Gradle plugin that breaks the build when the latest version of Spring Cloud Contract is used:

- `targetFramework` was deprecated attribute and it is now removed. This causes the Gradle configuration script to break and not compile
- Since Spring Cloud Contract Gradle plugin 2.2.0.M1 the generated tests are executed in separate task - `contractTest`. It need additional configuration in order to run the generated tests.

plamentotev:
The currently generated configuration for the Spring Cloud Contract Gradle plugin does not work with the latest version of the plugin.

 - `targetFramework` was deprecated attribute and it is now deleted. As the new attribute (`testFramework`) was available since at least Spring Boot 2.2.x use it instead so the configuration can work with all version after Spring Boot 2.2.x
 - Since Spring Cloud Contract Gradle plugin 2.2.0.M1 the generated tests are executed in separate task - `contractTest`. This tasks also need to be configured to use JUnit Platform (JUnit 5). Otherwise the generated tests are not going to be run.
    
#296 mentions that after JUnit 5 is made default for Spring Cloud Contract Gradle plugin the configuration should be revisited. As JUnit 5 is now the default (see spring-cloud/spring-cloud-contract#1249) please advice how to proceed. Change the configuration so that the `testFramework` is added only for certain Spring Boot version range?

Fixes #612

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractGradleBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractGradleBuildCustomizer.java
@@ -41,6 +41,8 @@ class SpringCloudContractGradleBuildCustomizer implements BuildCustomizer<Gradle
 
 	private static final VersionRange SPRING_BOOT_2_2_OR_LATER = VersionParser.DEFAULT.parseRange("2.2.0.M1");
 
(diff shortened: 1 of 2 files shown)


snicoll:
@plamentotev thank you very much for reporting the issue and for the PR.

PR #636, opened 2021-02-03
Updating Spring Cloud release train to 2020.0.1

marcingrzejszczak:
is this ok @OlgaMaciaszek @spencergibb @ryanjbaxter ?

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -83,11 +83,14 @@ initializr:
           - compatibilityRange: "[2.4.0.M2,2.4.0-M3]"
             version: 2020.0.0-M4
             repositories: spring-milestones
-          - compatibilityRange: "[2.4.0.M4,2.4.x-SNAPSHOT)"
+          - compatibilityRange: "2.4.0.M4"
             version: 2020.0.0
             repositories: spring-milestones
+          - compatibilityRange: "[2.4.0,2.5.0-M1)"
+            version: 2020.0.1
+            repositories: spring-milestones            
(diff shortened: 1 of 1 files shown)


ryanjbaxter (approved):

start-site/src/main/resources/application.yml
```diff
@@ -83,11 +83,14 @@ initializr:
           - compatibilityRange: "[2.4.0.M2,2.4.0-M3]"
             version: 2020.0.0-M4
             repositories: spring-milestones
-          - compatibilityRange: "[2.4.0.M4,2.4.x-SNAPSHOT)"
+          - compatibilityRange: "2.4.0.M4"
```
wilkinsona, line 86: This doesn't look quite right. `2.4.0.M4` means `2.4.0.M4` and later. As such it overlaps with `[2.4.0,2.5.0-M1)` below.

wilkinsona (changes requested):

start-site/src/main/resources/application.yml
marcingrzejszczak: ah, I thought it means `2.4.0.M4` only...

marcingrzejszczak:
So hopefully after the last commit:

- 2020.0.0 will be applied to anything from 2.4.0.M4 to 2.4.0 (inclusive)
- 2020.0.1 will be applied to anything from 2.4.1 on.
- 2020.0.2-SNAPSHOT will be applied for 2.4.x snapshots
- we don't support 2.5.x snapshots

wilkinsona (approved):

PR #647, opened 2021-02-25
Add additional reference links for Spring Cloud config clients

OlgaMaciaszek:
Add reference links related to external config imports to `HELP.md` for SC Config Client projects.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -1128,14 +1128,8 @@ initializr:
           artifactId: spring-cloud-starter-config
           links:
             - rel: reference
-              href: https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_spring_cloud_config_client
+              href: https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_client_side_usage
               description: Config Server Client
-            - rel: reference
-              href: https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#config-data-import
-              description: SpringBoot Config Data Import
-            - rel: reference
-              href: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config-files
-              description: Spring Boot External Application Properties
         - name: Config Server
           id: cloud-config-server
           description: Central management for configuration via Git, SVN, or HashiCorp Vault.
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -1036,6 +1036,16 @@ initializr:
...
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#_spring_cloud_config_client
+              description: Config Server Client
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-config/docs/current/reference/html/#config-data-import
+              description: SpringBoot Config Data Import
+            - rel: reference
+              href: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config-files
+              description: Spring Boot External Application Properties
```
snicoll, line 1048: Do we really need three links here? Links are rendered in a bullet list now so I'd like us to be conservative. 

Along the same lines of our discussion about the default value, if we believe this is an important part of the getting started experience, I'd rather have a section in the reference guide that covers that, eventually linking (with more text) to other part of the reference guide, including that link in Spring Boot, rather than a flat list like this.

start-site/src/main/resources/application.yml
```diff
@@ -1050,16 +1060,46 @@ initializr:
           description: Provides client-side support for externalized configuration in a distributed system. Using HashiCorp's Vault you have a central place to manage external secret properties for applications across all environments.
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-vault-config
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-cloud-vault/docs/current/reference/html/#_quick_start
```
snicoll, line 1065: There's an extra `_` in the anchor that makes this link invalid.

snicoll: So that quickstart is exactly what I had in mind. If you think the two other links are important, I'd rather update that section to add them and add only this link here.

snicoll (changes requested):
Thanks for the PR. I'd generally favor improving the reference guide (if necessary) and points to a "quickstart" section rather than 3 isolated links that users have to consult to patch things up themselves.

OlgaMaciaszek:
To me, it seems useful to add various bullet points linking to relevant documentation parts for the users to get at hand. Since you prefer adding less information, I would add a paragraph about it to the Quickstart section and remove some of the links. Still, since these starters usually correspond to particular tools within larger projects, I think we will need at least 2 links for all those dependencies, i.e.:
- project quickstart section link (for example, `quickstart` for Spring Cloud Config project)
- tool/ starter dedicated reference section link (for example, link to Config Client section within Spring Cloud Config project documentation)

OlgaMaciaszek:
As a different, I might also add the warning in the introductory part of the section dedicated to the tool/starter, for example, Config Client and not in the quickstart, and then will have one link only; wdyt?

OlgaMaciaszek:
@snicoll I have modified/ added the quickstart sections for the relevant projects. Once the changes are released next week, will commit relevant changes here.

PR #651, opened 2021-03-05
Upgrade to Spring Cloud Open Service Broker 3.3.0

royclarkson:


start-site/src/main/resources/application.yml
```diff
@@ -1017,16 +1017,16 @@ initializr:
...
-          compatibilityRange: "[2.0.0.RELEASE,2.4.0-M1)"
+          compatibilityRange: "[2.0.0.RELEASE,2.5.0-M1)"
           description: Framework for building Spring Boot apps that implement the Open Service Broker API, which can deliver services to applications running within cloud native platforms such as Cloud Foundry, Kubernetes and OpenShift.
           groupId: org.springframework.cloud
           artifactId: spring-cloud-starter-open-service-broker
           mappings:
-            - compatibilityRange: "[2.2.0.RELEASE,2.4.0-M1)"
-              version: 3.1.1.RELEASE
+            - compatibilityRange: "[2.2.0.RELEASE,2.5.0-M1)"
+              version: 3.3.0
```
snicoll, line 1026: Spring Cloud Open Service Broker `3..1.x` builds against Spring  Boot `2.2.x` while this new `3.3.0` release builds against Spring Boot `2.4.x`. 

IMO, there is no need to have such a wide range of Spring Boot versions for this. Rather, I'd add a new mapping entry as follows:

```yaml
  - compatibilityRange: "[2.4.0-M1,2.5.0-M1)"
    version: 3.3.0
```

snicoll (commented):
Thank you for keeping us current. I am not sure the mapping should be changed as you suggest, can you please have a look?

start-site/src/main/resources/application.yml
royclarkson: Sure, I can add the older versions back. You know, I even considered that you would probably ask me that exact question. :)

royclarkson: I've updated the PR.

snicoll:
Thank you for the follow-up @royclarkson. I've polished the version ranges in c6a977a.

PR #710, opened 2021-06-10
Upgrade to Vaadin 14.6.3

ZheSun88:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -150,8 +150,8 @@ initializr:
         artifactId: vaadin-bom
         versionProperty: vaadin.version
         mappings:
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -150,8 +150,8 @@ initializr:
         artifactId: vaadin-bom
         versionProperty: vaadin.version
         mappings:
-          - compatibilityRange: "[2.1.0.RELEASE,2.5.0-M1)"
-            version: 14.6.2
+          - compatibilityRange: "[2.1.0.RELEASE,2.5.0]"
```
snicoll, line 153: Why was the range here changed?

snicoll:
As discussed, we're going to open the 2.5.x range for this version as well.

PR #711, opened 2021-06-11
Improve the description of the Vaadin entry

mstahv:
Web components are there, but more like an implementation detail from Vaadin Flow users point of view.


snicoll:
Cool, thanks for reviewing!

PR #741, opened 2021-07-30
Upgrade to Spring Cloud 2021.0.0-M1

ryanjbaxter:
Adding support for 2021.0.0-M1

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -86,6 +86,12 @@ initializr:
         mappings:
           - compatibilityRange: "[2.2.0.RELEASE,2.4.0.M1)"
             version: Hoxton.SR12
+          - compatibilityRange: "[2.4.0.M1,2.4.0.M1]"
+            version: 2020.0.0-M3
+            repositories: spring-milestones
+          - compatibilityRange: "[2.4.0.M2,2.4.0-M3]"
+            version: 2020.0.0-M4
+            repositories: spring-milestones
           - compatibilityRange: "[2.4.0.M4,2.4.0]"
             version: 2020.0.0
           - compatibilityRange: "[2.4.1,2.5.x-SNAPSHOT)"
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -86,19 +86,19 @@ initializr:
         mappings:
           - compatibilityRange: "[2.2.0.RELEASE,2.4.0.M1)"
             version: Hoxton.SR12
-          - compatibilityRange: "[2.4.0.M1,2.4.0.M1]"
-            version: 2020.0.0-M3
-            repositories: spring-milestones
-          - compatibilityRange: "[2.4.0.M2,2.4.0-M3]"
-            version: 2020.0.0-M4
-            repositories: spring-milestones
```
snicoll, line 94: We can prune older metadata but I am not keen to do this as par of adding `2021.0.0-M1`.

start-site/src/main/resources/application.yml
```diff
@@ -86,19 +86,19 @@ initializr:
...
           - compatibilityRange: "[2.4.0.M4,2.4.0]"
             version: 2020.0.0
           - compatibilityRange: "[2.4.1,2.5.x-SNAPSHOT)"
             version: 2020.0.3
           - compatibilityRange: "2.5.x-SNAPSHOT"
             version: 2020.0.4-SNAPSHOT
             repositories: spring-snapshots,spring-milestones
+          - compatibilityRange: "[2.6.0.M1,2.6.0.M1]"
```
snicoll, line 96: Compatibility ranges must cover a supported generation. An easier way to write this would be to limit the range to `2.6.0-SNAPSHOT` so that the next entry covers the remaining of `2.6.x`.  `"[2.6.0.M1,2.6.x-SNAPSHOT)"` should do it

snicoll (changes requested):
Thanks for the PR. I'd like a review on the compatibility range on the new entry.

start-site/src/main/resources/application.yml
ryanjbaxter: OK should I submit a separate PR to remove these then?

start-site/src/main/resources/application.yml
ryanjbaxter: Thanks

start-site/src/main/resources/application.yml
snicoll: No need to, I can take care of that.

ryanjbaxter: Thanks!

PR #743, opened 2021-08-06
Upgrade to Spring Boot Admin 2.4.3

SteKoe:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -51,8 +51,8 @@ initializr:
         artifactId: spring-boot-admin-dependencies
         versionProperty: spring-boot-admin.version
         mappings:
-          - compatibilityRange: "[2.3.0.M1,2.5.0-M1)"
-            version: 2.3.1
+          - compatibilityRange: "[2.3.0.M1,2.6.0-M1)"
+            version: 2.4.3
       solace-spring-boot:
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -51,8 +51,8 @@ initializr:
         artifactId: spring-boot-admin-dependencies
         versionProperty: spring-boot-admin.version
         mappings:
-          - compatibilityRange: "[2.3.0.M1,2.5.0-M1)"
```
snicoll, line 54: Shouldn't we keep the `2.3.x` version for Spring Boot < 2.4.x ?

snicoll (commented):
Thanks for the PR. I am wondering about supporting a wide range like this. Shouldn't we keep the old range and add a new one instead?

start-site/src/main/resources/application.yml
SteKoe: Hi @snicoll! I have talked to the development team and we agree that the version range I have defined in this PR is correct since Spring Boot Admin is downwards compatible. :)

snicoll: Fair enough.

snicoll:
Thanks @SteKoe, I've polished your contribution in ecf4c8c.

Issue #760, opened 2021-09-21
Add support for Java 17


joshlong:
Java 17 is an LTE release and generally available from a number of vendors / distributions including Java.NET (OpenJDK), Azul, Liberica, AWS Coretto, etc. Thank you for your consideration and thank you for helping move the java ecosystem forward by proactively supporting the next LTE version of Java. this is also fruitful because Java 16 is technology no longer supported, anyway

snicoll:
The plan is to add Java 17 as soon as we have a Spring Boot release that officially supports it. This is happening this Thursday so I'll prepare the ground work to be ready on time. Thanks for raising it!

PR #761, opened 2021-09-21
Java 17

joshlong:
Please support Java 17 as per #760


snicoll:
Thank you for the PR but there are more things to do to support it, such as downgrading to Java 11 when we detect that the user has chosen an unsupported option (i.e. Kotlin).

PR #770, opened 2021-10-06
Upgrade to Native Buildtools 0.9.5

bclozel:
Prior to this commit, the GraalVM Native Gradle plugin would be added
and configured automatically for applications requiring the Spring
Native dependency.

After Spring Native 0.11.0-M1, this plugin is automatically added and
configured by the AOT plugin so additional customization is not needed
anymore.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springnative/SpringNativeBuildtoolsVersionResolver.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springnative/SpringNativeBuildtoolsVersionResolver.java
@@ -31,7 +31,9 @@ import io.spring.initializr.generator.version.VersionRange;
 abstract class SpringNativeBuildtoolsVersionResolver {
 
 	private static final List<NativeBuildtoolsRange> ranges = Arrays.asList(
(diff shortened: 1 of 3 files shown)


start-site/src/main/java/io/spring/start/site/extension/dependency/springnative/SpringNativeBuildtoolsVersionResolver.java
```diff
@@ -31,7 +31,9 @@
 abstract class SpringNativeBuildtoolsVersionResolver {
 
 	private static final List<NativeBuildtoolsRange> ranges = Arrays.asList(
-			new NativeBuildtoolsRange("[0.9.0,0.10.0-M1)", null), new NativeBuildtoolsRange("0.10.0-M1", "0.9.3"));
+			new NativeBuildtoolsRange("[0.9.0,0.10.0-M1)", null),
+			new NativeBuildtoolsRange("[0.10.0-M1, 0.11.0-M2)", "0.9.3"),
+			new NativeBuildtoolsRange("[0.11.0-M2,0.11.0-SNAPSHOT)", null));
```
snicoll, line 36: There are two features into one here, isn't it? `0.10.0-SNAPSHOT` also upgraded to Spring Native Buildtoold `0.9.5`. I understand that the Gradle plugin takes care of this automatically and that's great, but isn't the Maven side of things affected as well?

PR #793, opened 2021-11-16
Upgrade to Spring Cloud Services 3.3.0

kvmw:
Spring cloud services is compatible with spring boot 2.5.x.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -112,6 +112,8 @@ initializr:
             version: 2.3.0.RELEASE
           - compatibilityRange: "[2.4.0-M1,2.5.0-M1)"
             version: 2.4.1
+          - compatibilityRange: "[2.5.0-M1,2.6.0-M1)"
+            version: 3.3.0          
       spring-geode:
         groupId: org.springframework.geode
         artifactId: spring-geode-bom
@@ -1274,7 +1276,7 @@ initializr:
           artifactId: spring-cloud-stream
     - name: VMware Tanzu Application Service
       bom: spring-cloud-services
-      compatibilityRange: "[2.0.0.RELEASE,2.5.0-M1)"
+      compatibilityRange: "[2.0.0.RELEASE,2.6.0-M1)"
       content:
         - name: Config Client (TAS)
           id: scs-config-client
@@ -1285,7 +1287,7 @@ initializr:
             - rel: reference
               href: https://docs.pivotal.io/spring-cloud-services/
           mappings:
-            - compatibilityRange: "[2.0.0.RELEASE,2.4.0-M1)"
+            - compatibilityRange: "[2.0.0.RELEASE,2.6.0-M1)"
               starter: false
         - name: Service Registry (TAS)
           id: scs-service-registry


start-site/src/main/resources/application.yml
```diff
@@ -1274,7 +1274,7 @@ initializr:
           artifactId: spring-cloud-stream
     - name: VMware Tanzu Application Service
       bom: spring-cloud-services
-      compatibilityRange: "[2.0.0.RELEASE,2.5.0-M1)"
+      compatibilityRange: "[2.0.0.RELEASE,2.6.0-RC1)"
```
snicoll, line 1277: I don't understand the range. Did you mean `2.6.0-M1` ?

start-site/src/main/resources/application.yml
```diff
@@ -1285,7 +1285,7 @@ initializr:
             - rel: reference
               href: https://docs.pivotal.io/spring-cloud-services/
           mappings:
-            - compatibilityRange: "[2.0.0.RELEASE,2.4.0-M1)"
+            - compatibilityRange: "[2.0.0.RELEASE,2.6.0-RC1)"
```
snicoll, line 1288: Ditto

snicoll (changes requested):
Thanks for the PR but the ranges don't look right to me. It looks like `3.3.0` should be used for Spring Boot 2.5.x but I don't see it in this PR. The range for the bom looks wrong as well.

I can take over if you like but I'd rather prefer a description of what's expected from the SCS team as it enforces the decision on the team.

start-site/src/main/resources/application.yml
kvmw: It was meant to be any version of 2.5.x (but not 2.6).  So I used the first RC version of 2.6 (exclusive).

kvmw:
@snicoll , 

> Thanks for the PR but the ranges don't look right to me. It looks like `3.3.0` should be used for Spring Boot 2.5.x but I don't see it in this PR. The range for the bom looks wrong as well.
> 
> I can take over if you like but I'd rather prefer a description of what's expected from the SCS team as it enforces the decision on the team.

@snicoll , 
We just realised that spring cloud services dependencies are disabled for boot 2.5.x in initializer, because of the compatibility range (currently set to 2.5.0-M1).  
This was supposed to be a simple change to enabled  SCS dependencies ( Client Config, Service Registry and Circuit Breaker) for boot 2.5.x in intializer.

start-site/src/main/resources/application.yml
snicoll: `2.6.0-RC1` is not the first version in the 2.6.x line so that change would mean any milestone of 2.6.x would be included (`2.6.0-M1`, `2.6.0.M2`, etc. up to `2.6.0-RC1` excluded).

snicoll:
> This was supposed to be a simple change to enabled SCS dependencies ( Client Config, Service Registry and Circuit Breaker) for boot 2.5.x in intializer.

That's why we prefer that we describe the expectations rather than trying to figure out from the PR what you meant. That still doesn't answer my question about `3.3.0`. The [https://github.com/pivotal-cf/spring-cloud-services-starters/releases] release notes indicate that `3.3.0` is the one compatible with Spring Boot 2.5.x and this PR doesn't define it at all.

start-site/src/main/resources/application.yml
kvmw: sorry, my bad.

kvmw:
> > This was supposed to be a simple change to enabled SCS dependencies ( Client Config, Service Registry and Circuit Breaker) for boot 2.5.x in intializer.
> 
> That's why we prefer that you describe the expectations rather than trying to figure out from the PR what you meant. That still doesn't answer my question about `3.3.0`. The [release notes](https://github.com/pivotal-cf/spring-cloud-services-starters/releases) indicate that `3.3.0` is the one compatible with Spring Boot 2.5.x and this PR doesn't define it at all.

I'm afraid i don't have enough context to answer your question about 3.3.0 release. I'll get back to you once i've more info.

snicoll:
This broke our metadata tests: https://ci.spring.io/teams/initializr/pipelines/start-site/jobs/verify/builds/1794#L61857ad2:437

kvmw:
> This broke our metadata tests: https://ci.spring.io/teams/initializr/pipelines/start-site/jobs/verify/builds/1794#L61857ad2:437

Sorry about that.  since version 3.0.0,  circuit breaker has been dropped. I'll create another PR for this today.

snicoll:
> Sorry about that. since version 3.0.0, circuit breaker has been dropped. I'll create another PR for this today.

I see. I have pushed something to fix our metadata test so no need for a PR, thanks.

PR #804, opened 2021-11-29
Update Vaadin compatibilityRange

tarekoraby:
Update Vaadin's compatibilityRange to indicate that Spring Boot `SNAPSHOT` versions are compatible with Vaadin version `21.0.8`.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -142,6 +142,10 @@ initializr:
         mappings:
           - compatibilityRange: "[2.1.0.RELEASE,2.6.0-M1)"
             version: 14.7.5
+          - compatibilityRange: "2.5.x-SNAPSHOT"
+            version: 21.0.8
+          - compatibilityRange: "2.6.x-SNAPSHOT"
+            version: 21.0.8
       wavefront:
         groupId: com.wavefront
         artifactId: wavefront-spring-boot-bom


start-site/src/main/resources/application.yml
```diff
@@ -142,6 +142,10 @@ initializr:
         mappings:
           - compatibilityRange: "[2.1.0.RELEASE,2.6.0-M1)"
             version: 14.7.5
+          - compatibilityRange: "2.5.x-SNAPSHOT"
```
snicoll, line 145: All Spring Boot 2.1.x to 2.6.0-M1 excluded use `14.7.5`.  As such, 2.5.x-SNAPSHOT are also using this version. I believe this range is inaccurate.

start-site/src/main/resources/application.yml
```diff
@@ -142,6 +142,10 @@ initializr:
...
+            version: 21.0.8
+          - compatibilityRange: "2.6.x-SNAPSHOT"
```
snicoll, line 147: What is the intent here. Offer this version as of Spring Boot 2.6? If so, it should use the same format as above, something like `[2.6.0-M1,2.7.0-M1)`.

Is that version of the Vaadin Spring integration built and tested against Spring Boot 2.6.x?

snicoll (commented):
Thanks for the PR. I am not sure I understood the intent so I've added a couple of comments.

start-site/src/main/resources/application.yml
tarekoraby: Not quite. The intent is to ship Vaadin v21x with Spring Boot SNAPSHOT only, while Spring Boot stable versions (i.e. 2.5.7 and 2.6.0) would come with Vaadin v14 which is the current LTS (I realize now that the `compatibilityRange` should be further adjusted so that Vaadin 14 would work with 2.6.0). 

The motivation is to get people to try Spring Boot SNAPSHOT + Vaadin vLatest combo so that we can catch any possible issues as early as possible.

snicoll: @tarekoraby that doesn't sound right to me.

Let's assume we merge what you described. `2.6.0` is using the LTS version and `2.6.1-SNAPSHOT` is using the non-LTS version (going 14.x to 21.x!). We're releasing `2.6.1` today. When that's done, `2.6.1` is using the LTS version again.

I am all for letting people try out the next version when they choose to use the SNAPSHOT, but going from 14.x to 21.x isn't reasonable IMO.

tarekoraby: I see you're point (and it's a valid one). Vaadin LTS still awaits an upcoming fix that should make it compatible Spring Boot 2.6.x. Once that arrives, let's make Vaadin LTS available for all the currently selectable Spring Boot versions (which would mean `compatibilityRange: "[2.1.0.RELEASE,2.6.2-SNAPSHOT]"` if I'm not mistaken.

The fix to Vaadin LTS should arrive next week at the latest. Should I close this PR for now and open a new one then?

snicoll:
Let's close this one per https://github.com/spring-io/start.spring.io/pull/804#discussion_r759086114. For the continuity argument I mentioned above, the next LTS should only be provided as of Spring Boot 2.6. You can provide a PR or describe your needs in a separate issue once the 2.6.x compatibility issues are  fixed. Thank you!

PR #806, opened 2021-11-30
Upgrade to Vaadin 14.7.7

ZheSun88:
with this version, vaadin is compatible with spring boot 2.6.x

let us have a separate PR for the compatibility range discussion.


snicoll:
@ZheSun88 thanks, I did upgrade the range as part as merging this.

PR #814, opened 2021-12-18
Upgrade to Spring Boot Admin 2.5.5

SteKoe:
2.5.5 includes log4j and thymeleaf updates to mitigate CVE issues


snicoll:
Thanks @SteKoe. Any news of a version compatible with Spring Boot 2.6?

snicoll:
@SteKoe thanks for keeping us current. I believe the range change was incorrect so I've fixed that in 0ff81ad

SteKoe:
@snicoll thanks fixing the Version range issue. We started to work on Spring Boot 2.6 support today! :)

PR #827, opened 2022-01-14
Upgrade to Spring Cloud GCP 3.0.0

elefeint:
Spring Cloud GCP 3.0.0 adds compatibility with Spring Boot 2.6 and Spring Cloud 2021.0.0.

Other changes:
* Removed per-artifact compatibility mappings for GCP modules in favor of pre-existing global `compatibilityRange`. The `spring-cloud-gcp` BOM definition should take care of versioning.
* Updated outdated documentation links.

I've tested these changes locally, and they seem to work well for both, Spring Boot 2.5.x and 2.6.x. But any pointers for what this configuration should ideally look like would be very welcome.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -108,6 +108,8 @@ initializr:
         mappings:
           - compatibilityRange: "[2.4.0-M1,2.6.0-M1)"
             version: 2.0.7
+          - compatibilityRange: "[2.6.0,2.7.0-M1)"
+            version: 3.0.0
       spring-cloud-services:
         groupId: io.pivotal.spring.cloud
         artifactId: spring-cloud-services-dependencies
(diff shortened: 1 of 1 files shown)


elefeint:
@meltsufin FYI.

snicoll:
@elefeint thanks for asking and keeping us current.

We usually keep support for versions that are not EOL yet, albeit not being offered as an option. It is still possible to create a `2.3.x` or `2.4.x` projects using the CLI, and we like those users to be able to get the right version for a dependency. That said, I noticed that the bom mapping was for 2.4.x onwards so I've fixed the compatibility range instead.

Speaking of such a range, they need to overlap so I've changed the `2.6.0` lower bound to `2.6.0-M1`.  You can find all this in https://github.com/spring-io/start.spring.io/commit/aa37a7a9a791f02ff129651196f1b927831074c9. Thanks again!

PR #849, opened 2022-02-11
Add tomcat-embed-programmatic dependency when web and native are selected

eddumelendez:
Spring Native documentation mention:
> org.apache.tomcat.experimental:tomcat-embed-programmatic dependency should be used for optimized footprint.


snicoll:
Thank you for the PR but I am not keen to do this at this time. There is some work stream where this shouldn't be necessary in the future and I prefer the build to stay as idiomatic as possible.

PR #861, opened 2022-03-02
Add support for Vaadin 23.0.0

ZheSun88:
We will release Vaadin 23.0.0 today, which is the latest LTS version by us. 
and at the same time, V14 is still supported until 2023(roughly)

I am a bit uncertain whether the compatibility range can be overlapped between different versions, 
so please let me know if this is doable, thanks

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -156,9 +156,9 @@ initializr:
         artifactId: vaadin-bom
         versionProperty: vaadin.version
         mappings:
-          - compatibilityRange: "[2.1.0.RELEASE,2.8.0-M1)"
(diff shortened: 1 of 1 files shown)


snicoll:
@ZheSun88 thanks for the PR and for keeping us current. No, the range can't overlap. What would this mean anyway. This PR states that two versions can be used as of `2.6.4` so which one should we pick?

ZheSun88:
Hi@snicoll , thanks for the quick reply. 
just discussed with our dev team and did some tests. I have updated the version range. 
we'd like to let the user to get V23.0.0 when they choose spring-boot 2.6.X (2.6.4 latest). 

let me know if this is okay. Thanks again.

snicoll:
@ZheSun88 thanks for the update and congrats on the new LTS

Issue #783, opened 2021-10-29
Missing Amazon Web Services entry


vojtapol:
The [documentation for the Spring Cloud for AWS](https://spring.io/projects/spring-cloud-aws) states that it is part of the Spring Cloud project:
> Spring Cloud for Amazon Web Services, part of the Spring Cloud umbrella project...

It also has a banner that leads to Spring Initializr:
<img width="818" alt="image" src="https://user-images.githubusercontent.com/7419355/139361982-07662b35-cdc7-4f19-bf0f-e2042735a668.png">

However I can't seem to be able to add a dependency to any of the Spring Cloud AWS projects using the Spring Iniitializr user interface. Specifically, I am trying to figure out how to add dependency on `org.springframework.cloud:spring-cloud-starter-aws-jdbc` without explicitly specifying its version (I want to inherit it from BOM or parent). I was hoping that Spring Initializr would help me declare the dependencies correctly.

I don't understand why is it not there or whether it will be ever added?

If it cannot be or will not be added, should the documentation be updated to not lead to the Spring Initializr site?

PR #868, opened 2022-03-10
Adding Amazon Web Services (AWS) back

deki:
I noticed the AWS dependencies had been removed as part of commit cd6d17b1 as Spring Boot 2.3.x is EOL. This had already been reported in #783. Newer versions supporting Spring Boot >2.3.x are available but the groupId changed.


snicoll:
@deki thanks for the PR but unless you are working on Spring Cloud AWS, this isn't what we need. The reason why it was removed (as explained in the issue) is the lack of timely updates by the team managing the project. Before we consider re-adding them, we'd need to review the entries, see if they are still making sense and validate them. 

If you are not working on Spring Cloud AWS, please report that issue to them.

maciejwalkowiak:
At this stage we prefer to stay out of start.spring.io as we cannot guarantee updates exactly on time as @snicoll mentioned.

snicoll:
Thanks for the follow-up @maciejwalkowiak

Issue #460, opened 2020-06-04
Links in HELP.md should consistently link to either the single-page documentation or the multi-page documentation


wilkinsona:
The links that we add to HELP.md are inconsistent in the reference documentation that they link to. Some point to `html` and others to `htmlsingle`. For example, these are the links from an app I generated recently:

```markdown
* [Official Apache Maven documentation](https://maven.apache.org/guides/index.html)
* [Spring Boot Maven Plugin Reference Guide](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/maven-plugin/reference/html/)
* [Create an OCI image](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/maven-plugin/reference/html/#build-image)
* [Prometheus](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/reference/html/production-ready-features.html#production-ready-metrics-export-prometheus)
* [Spring Boot Actuator](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/reference/htmlsingle/#production-ready)
* [Spring Web](https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/reference/htmlsingle/#boot-features-developing-web-applications)
```

Related to this, `htmlsingle` links with an anchor, such as https://docs.spring.io/spring-boot/docs/2.3.0.RELEASE/reference/htmlsingle/#production-ready, do not work in Eclipse. When previewing HELP.md, they are rendered as links, but clicking on them does nothing. For this reason, in the absence of any strong reason to do something else, I'm tempted to settle on using `html` rather than `htmlsingle`.

PR #869, opened 2022-03-17
Review metadata and dependency links

robertmcnees:
Updated and verified all links for dependencies.


robertmcnees:
Possible fix for #460 and #79

PR #885, opened 2022-04-01
Upgrade to Spring Cloud GCP 2.0.9 and 3.2.0

elefeint:
We've released Spring Cloud GCP 2.0.9 and 3.2.0.


snicoll:
@elefeint thank you for keeping us current!

PR #886, opened 2022-04-01
Upgrade to Spring Cloud 2022.0.0-M2

ryanjbaxter:


snicoll:
Thanks @ryanjbaxter but we don't want to remove the metadata for older Spring Boot versions. I've upgraded in [0e37692](https://github.com/spring-io/start.spring.io/commit/0e376927b220517a7b22a4fa564be6fd685522b4)

PR #945, opened 2022-06-18
Fix link to "Building REST services with Spring"

xnumad:
Currently returns 404


xnumad:
@pivotal-cla This is an Obvious Fix

snicoll:
@xnumad thanks for the PR but the API mock file you've changed is a test resource for the client. You are right though that the link is broken. 

The correct location is https://github.com/spring-io/start.spring.io/blob/215bd0d192a0ee8efe74ba478896171fd00c2836/start-site/src/main/resources/application.yml#L323

Issue #658, opened 2021-03-12
Provide error pages


snicoll:
Now that we've re-enabled the whitelabel page, we should provide a static error page that's a bit more inline with the current UI.

10xtechie:
@snicoll ,
Do you mean that we should have a static error page e.g `error.html` and fill it with appropriate HTTP error code such as `404`, `500` etc - with a generic error massage? Also, do you want any further details about the error; for example: which java exception causes the `500` error, so on and so forth?

10xtechie:
### @snicoll 
Thanks so very much for your reply!!
Let me think aloud here: on the sever side, I would create a generic error handling controller which will produce usual error attributes both in HTML and JSON and on the client side, a generic error page template, which is consistent with our web UI and will consume data produced by that generic error controller. Please let me know what you think :)

snicoll:
> I would create a generic error handling controller 

That is supposed to be taken care by Spring Boot's default error handling. Unfortunately this proves to be a bit tricky with our web client for some reason. I am happy for you giving a try. We can discuss based on a PR.

10xtechie:
@snicoll ,
Thanks for insights! I will try and create a PR. Please assign it to me :)

10xtechie:
@snicoll ,
Thanks for assigning this issue to me! I have created a [PR ](https://github.com/spring-io/start.spring.io/pull/961)for the same. Please have a look.

PR #961, opened 2022-07-07
issue #658 - initial changes for adding error pages for both client a…

10xtechie:
Initial changes for adding custom error pages for client and messages for server


snicoll:
Thanks for the PR and sorry it took so long to process it. With the introduction of `ProblemDetails` in Spring Boot3, I'd like to explore that route rather than creating a custom `ErrorController`. I ran this locally and couldn't manage to make it work.

PR #962, opened 2022-07-10
Add flyway dependencies for mariadb, mysql, sqlserver

eddumelendez:
Spring Boot 2.7.x provides dependencies for `flyway-core`,
`flyway-mysql` and `flyway-sqlserver` which are needed when
flyway and the database driver are present.


snicoll:
Nice, thanks @eddumelendez

PR #982, opened 2022-08-12
Moved goal prepare-frontend of the Vaadin plugin to the build

simasch:
The goal prepare-frontend must run with the build no matter which profile is activated to make the Vaaddin application run otherwise the Vaadin application does not start.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/vaadin/VaadinMavenBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/vaadin/VaadinMavenBuildCustomizer.java
@@ -29,9 +29,12 @@ class VaadinMavenBuildCustomizer implements BuildCustomizer<MavenBuild> {
 
 	@Override
 	public void customize(MavenBuild build) {
+		build.plugins().add("com.vaadin", "vaadin-maven-plugin", (plugin) -> plugin.version("${vaadin.version}")
+				.execution("prepare-frontend", (execution) -> execution.goal("prepare-frontend")));
+
 		build.profiles().id("production").plugins().add("com.vaadin", "vaadin-maven-plugin",
 				(plugin) -> plugin.version("${vaadin.version}").execution("frontend",
-						(execution) -> execution.goal("prepare-frontend").goal("build-frontend").phase("compile")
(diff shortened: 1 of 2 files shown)


simasch:
According to my tests, prepare-frontend seems not necessary. I have to check with the Vaadin guys.

simasch:
After discussing with the Vaadin team I reopen the PR.
All examples and also projects created on https://start.vaadin.com contain the Vaadin plugin in the build section.

Sorry about the confusion

mstahv:
I was part of that discussion and I'd leave it like this still, at least for a while.

 * Vaadin + Spring Boot projects don't really need that the "prepare-frontend" goal for "development mode", like certain servlet containers
 * In normal usage one shouldn't need the plugin, except when making a production build, that is triggered with a profile like today
 * In the current setup, the pom.xml is cleaner, which at least I find valuable (I hate to say it, but I hate what the pom.xml looks like when you create a project via the referred start.vaadin.com)
 * The only really good reason that I have heard to have it is that calling "mvn vaadin:clean-frontend" is easier than "mvn com.vaadin:vaadin-maven-plugin:LATEST:clean-frontend" or "mvn vaadin:clean-frontend -Pproduction". This goal provides additional cleanup for the front-end build done by Vaadin. Normal user should never need to use it (and calling it slows down the subsequent front-end build), but it is often used to resolve some bugs that render the webpack based build into some sort of non-functional state. Those issues have been real and clean-frontend goal has helped as a workaround to solve them, but the actual fixes ought to go to Vaadin. The soon to be released 23.2.0 comes with a Vite based build as a default, that makes me again hope that the clean-frontend goal would end up being a relic that reminds us why we want to stay out of the front-end business 🤐

Thus, I'd look into this again in couple of months and then re-consider this. If the need for clean-frontend shortcut is not critical anymore, we should probably instead use this current "Spring Initializr approach" also in the projects that are generated by Vaadin archetype and start.vaadin.com

snicoll:
Thanks for the insights @mstahv. The argument of having a clean and focused pom is really what we're trying to achieve here and I wanted to double check that we were not out-of-sync. 

@simasch thanks for the PR but we'll keep things as they are for now.

PR #983, opened 2022-08-18
Upgrade to Okta Spring Boot 2.1.6

arvindkrishnakumar-okta:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -522,7 +522,7 @@ initializr:
             - compatibilityRange: "[2.4.1,2.5.0-M1)"
               version: 2.0.1
             - compatibilityRange: "[2.5.0-M1,2.7.0-M1)"
-              version: 2.1.5
+              version: 2.1.6
           links:
             - rel: guide
               href: https://github.com/okta/samples-java-spring/tree/master/okta-hosted-login


start-site/src/main/resources/application.yml
```diff
@@ -522,7 +522,7 @@ initializr:
             - compatibilityRange: "[2.4.1,2.5.0-M1)"
               version: 2.0.1
             - compatibilityRange: "[2.5.0-M1,2.7.0-M1)"
```
mraible, line 524: I've had success using the Okta Spring Boot starter with Spring Boot 3. Is there any reason not to increase this compatibility range?

mraible:
@snicoll Can you please help with this?

start-site/src/main/resources/application.yml
snicoll: We don't generally think it's a good idea for a single version of a dependency to be use with both Spring Boot 2 and Spring Boot 3.

snicoll: However, it's a bit strange that the entry is still blocked for Spring Boot 2.7 that GA for months now...

snicoll: Given https://github.com/okta/okta-spring-boot/blob/7c9742fb5fee90c38fb2afc947f7b67047bb1dd6/pom.xml#L34 I am going to increase the range...

Issue #987, opened 2022-08-23
Add Spring Pulsar


onobc:
# Requirements

## Code of Conduct
This is a Spring project and as such follows the [Spring CoC](https://github.com/spring-projects-experimental/spring-pulsar/blob/main/CODE_OF_CONDUCT.adoc).

## Well-established project with a vibrant community
A goal of [Spring Pulsar](https://github.com/spring-projects-experimental/spring-pulsar) is to make it easy for Spring apps to use [Apache Pulsar](https://github.com/apache/pulsar) messaging. While Spring Pulsar is in the experimental stage, Apache Pulsar is a well-established project with a vibrant community. We want Spring Pulsar to ride that wave with Apache Pulsar. Adding support here could help users experiment w/ Spring Pulsar. 

_Also worth pointing out is that a Spring Tips video by Josh Long is slated to be released on **09/05/2022** and we would love to have this available for that video if possible._

## Ongoing maintenance
We do follow Spring Boot release and maintenance cadence and update accordingly and in a timely fashion.

## License
We use the same license as other Spring projects, [Apache License 2.0](https://github.com/spring-projects-experimental/spring-pulsar/blob/main/LICENSE.txt).

## Issue Tracker URL
We use Github for issue tracking: https://github.com/spring-projects-experimental/spring-pulsar/issues

## Continuous integration
We use Github actions for CI: https://github.com/spring-projects-experimental/spring-pulsar/actions

_**NOTE**: We only support Spring Boot 3.0.x currently and therefore do not have a compatibility CI job yet._

## Maven Central
This should be listed as "experimental" and will not be available yet on Maven Central but rather on [repo.spring.io](https://repo.spring.io/artifactory/libs-snapshot-local/org/springframework/pulsar/spring-pulsar/0.1.0-SNAPSHOT/)

## Configuration metadata
We do provide configuration properties and we do publish the annotation metadata with the starter. Additionally we dynamically generate the [config props documentation](https://docs.spring.io/spring-pulsar/docs/current-SNAPSHOT/reference/html/application-properties.html#appendix.application-properties).

## Version mappings

Only 1 mapping available:
* `SB 3.0.x -> spring-pulsar 0.1.x`

## Links to additional resources
* Reference documentation: https://docs.spring.io/spring-pulsar/docs/current-SNAPSHOT/reference/html/
* Sample: https://github.com/spring-projects-experimental/spring-pulsar/blob/main/spring-pulsar-sample-apps/README.adoc

[...]

PR #1002, opened 2022-09-30
Add spring pulsar

eddumelendez:
Spring for Apache Pulsar 0.1.0-M1 was released based on spring boot
3.0.0-M4. Also, if spring pulsar is selected testcontainers pulsar
module is added.


snicoll:
@eddumelendez I am afraid we don't accept PR for new entries. We haven't got the bandwidth to review #987 and that should happen first anyway.

Issue #1013, opened 2022-10-19
Use new coordinates for MySQL's Java Connector


wilkinsona:
Once Boot 2.6.12 and 2.7.5 have been released, we should update the MySQL dependency to use `com.mysql:mysql-connector-j`.

PR #1081, opened 2023-01-05
Fix MySQL coordinate errors

BingZi-233:


BingZi-233:
This is a very small change. I don't think it is necessary for me to sign the agreement.

wilkinsona:
Thanks for the PR but the coordinates are correct as they currently are. Please see https://github.com/spring-io/start.spring.io/issues/1013 for details.

BingZi-233:
Just now, I tried to obtain dependencies from the central warehouse. It seems that he is not working normally at present.

wilkinsona:
The artifact is available from Central: https://repo.maven.apache.org/maven2/com/mysql/mysql-connector-j/. Make sure you're using 8.0.31.

PR #1113, opened 2023-02-07
Add dev.hilla/hilla entry

ZheSun88:


ZheSun88:
We will start to support springboot 3.0.x with Vaadin 24 and Hilla 2.0.
They are currently in prerelease stage, which is in vaadin-prerelease repo instead of maven-central 
Can we add the support here?

wilkinsona:
See https://github.com/spring-io/start.spring.io/issues/872 to which this relates.

wilkinsona:
> They are currently in prerelease stage, which is in vaadin-prerelease repo instead of maven-central
Can we add the support here?

Not yet, I'm afraid. We require third-party projects to have been published to Maven Central before we add them to start.spring.io. Assuming that #872 is approved by the team, we could make the changes proposed here and then make a further change in the future use Hilla 2.0 has been released.

wilkinsona:
Thanks for the PR, @ZheSun88. In addition to the new bom, a new entry in the `dependencies` section is also needed. This would be similar to [what we already have for Vaadin](https://github.com/spring-io/start.spring.io/blob/e9a9ec07ca9ae3c6e13884e27d0e7c2a633ee7e0/start-site/src/main/resources/application.yml#L368-L382). Would you like to update the PR to propose something for this as well?

ZheSun88:
Now Hilla 2.0.0 has been released to maven. https://repo.maven.apache.org/maven2/dev/hilla/ 
@wilkinsona sure, i will take a look and add the new entry in the dependencies

ZheSun88:
added the hilla entry to dependency section

start-site/src/main/resources/application.yml
```diff
@@ -181,6 +181,13 @@ initializr:
             version: 23.2.15
           - compatibilityRange: "[2.7.0-M1,2.8.0-M1)"
             version: 23.3.5
+      hilla:
+        groupId: dev.hilla
+        artifactId: hilla-bom
+        versionProperty: hilla.version
+        mappings:
+          - compatibilityRange: "[2.7.0-M1,2.8.0-M1)"
```
wilkinsona, line 189: Unlike the range below (`[2.6.0-M1,3.1.0-M1)`), this range doesn't cover Spring Boot 3.0. I think another mapping is needed for the bom that's compatible with Spring Boot 3.

wilkinsona (changes requested):

start-site/src/main/resources/application.yml
ZheSun88: ah, indeed.. by that time, hilla 2.0 was not in maven.. i will update it.

ZheSun88:
Hi @wilkinsona , Thanks for the help. 

It seems we are missing some plugins in the pom.xml of the generated file.  Can you help us to add those?

there are two part of code
- add `hilla-maven-plugin` to pom 

```
            <build>
                <plugins>
                    <plugin>
                        <groupId>dev.hilla</groupId>
                        <artifactId>hilla-maven-plugin</artifactId>
                        <version>${hilla.version}</version>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>prepare-frontend</goal>
                                </goals>
                                <phase>compile</phase>
                            </execution>
                        </executions>
                    </plugin>
                </plugins>
            </build>
```

- add a `production` profile to pom.xml
```
<profile>
            <!-- Production mode is activated using -Pproduction -->
            <id>production</id>
            <dependencies>
                <!-- Exclude development dependencies from production -->
                <dependency>
                    <groupId>com.vaadin</groupId>
                    <artifactId>vaadin-core</artifactId>
                    <exclusions>
                        <exclusion>
                            <groupId>com.vaadin</groupId>
                            <artifactId>vaadin-dev</artifactId>
                        </exclusion>
                    </exclusions>
                </dependency>
            </dependencies>
            <build>
                <plugins>
                    <plugin>
                        <groupId>dev.hilla</groupId>
                        <artifactId>hilla-maven-plugin</artifactId>
                        <version>${hilla.version}</version>
                        <executions>
                            <execution>
                                <goals>
                                    <goal>build-frontend</goal>
                                </goals>
                                <phase>compile</phase>
                            </execution>
                        </executions>
                    </plugin>
                </plugins>
            </build>
        </profile>
```

wilkinsona:
@ZheSun88 Please open a new issue with some details about the desired Maven and Gradle configuration and we can take a look.

ZheSun88:
sorry for being a bit late.. here is the ticket. https://github.com/spring-io/start.spring.io/issues/1134

wilkinsona:
With some regret, we have reverted this change in 9509c65cdd16d61c723ce6e20fa5d745428d22cf. Unfortunately, when this was merged we did not realise that Hilla did not support both Maven and Gradle. We do not have the necessary infrastructure in place for a dependency to require a particular build system. Until that infrastructure is in place or Hilla supports both Maven and Gradle we don't think we're ready for it to be included on start.spring.io.

tarekoraby:
Thanks for the update, @wilkinsona. We are planning to add Gradle support to Hilla soon.

Issue #1144, opened 2023-03-24
Add Spring Authorization Server


wilkinsona:
Spring Boot 3.1.0-M2 has added a new starter, `spring-boot-starter-oauth2-authorization-server`, for Spring Authorization Server. We should offer a corresponding dependency on start.spring.io.

jongwooo:
Can I work on this issue?

sjohnr:
Hi @wilkinsona, I believe this should be unblocked now. Is it ready to be taken up?

sjohnr:
Thanks. @jongwooo did you still want to take this on? Otherwise, I'm happy to.

jongwooo:
@sjohnr Thank you for letting me know! Do you mind if I give it a try and let you know?

PR #1181, opened 2023-05-17
Add Spring Authorization Server

jongwooo:
Please see gh-1144

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -436,6 +436,15 @@ initializr:
               description: Authenticating a User with LDAP
             - rel: reference
               href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#web.security
+        - name: Spring Authorization Server
+          id: oauth2-authorization-server
+          description: Spring Boot integration for Spring Security's authorization server features.
+          facets:
+            - native
+          compatibilityRange: "3.1.0-RC1"
+          links:
+            - rel: reference
+              href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#web.security.oauth2.authorization-server
         - name: OAuth2 Client
           id: oauth2-client
           description: Spring Boot integration for Spring Security's OAuth2/OpenID Connect client features.


start-site/src/main/resources/application.yml
```diff
@@ -436,6 +436,14 @@ initializr:
               description: Authenticating a User with LDAP
             - rel: reference
               href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#web.security
+        - name: Spring Authorization Server
+          id: authorization-server
```
sjohnr, line 440: I'm assuming this would need to be `oauth2-authorization-server`?

start-site/src/main/resources/application.yml
```diff
@@ -436,6 +436,14 @@ initializr:
               description: Authenticating a User with LDAP
             - rel: reference
               href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#web.security
+        - name: Spring Authorization Server
+          id: oauth2-authorization-server
+          description: Spring Boot integration for Spring Security's authorization server features.
+          facets:
+            - native
+          links:
```
sjohnr, line 444: Support was added in Boot `3.1.0-M2` but really only worked since `3.1.0-RC1`. Should there be a `compatibilityRange` added?

jongwooo: > Support was added in Boot 3.1.0-M2 but really only worked since 3.1.0-RC1. 

@sjohnr I had concerns about adding `compatibilityRange`. Could you please let me know if there is a good way to add this?

wilkinsona: You can use `compatibilityRange: "3.1.0-RC1"` which means `3.1.0-RC1` and later.

sjohnr (approved):

start-site/src/main/resources/application.yml
```diff
@@ -436,11 +436,20 @@ initializr:
...
+            - rel: reference
+              href: https://docs.spring.io/spring-boot/docs/{bootVersion}/reference/htmlsingle/#web.security.oauth2.authorization-server
         - name: OAuth2 Client
           id: oauth2-client
           description: Spring Boot integration for Spring Security's OAuth2/OpenID Connect client features.
           facets:
             - native
+          compatibilityRange: 3.1.0-RC1
```
snicoll, line 452: It looks like the compatibility range was added on the wrong entry? Also please quote the value for consistency (see Andy's comment).

jongwooo: @snicoll I accidentally added it to the wrong dependency😅 Thank you for your help!

snicoll: Sure thing, thanks.

Issue #1190, opened 2023-05-24
Allow Vaadin 24 with Spring Boot 3.1


snicoll:
The Vaadin team would like us to open the range so that 3.1 is available as well.

PR #1191, opened 2023-05-24
Enable Vaadin with Spring Boot 3.1.0

cromoteca:
Vaadin 24.0 works fine with Spring Boot 3.1, so this PR increases the compatibility range for it.


snicoll:
Thanks but that's already done, see #1190. The build of that particular change failed due to a network glitch and I am looking to get that resolved

PR #1200, opened 2023-05-30
Upgrade Okta Spring Boot to 3.0.4

arvindkrishnakumar-okta:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -441,8 +441,8 @@ initializr:
               version: 2.0.1
             - compatibilityRange: "[2.5.0-M1,3.0.0-M1)"
               version: 2.1.6
-            - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
-              version: 3.0.3
+            - compatibilityRange: "[3.0.0-M1,3.1.0)"
+              version: 3.0.4
           links:
             - rel: guide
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -441,8 +441,8 @@ initializr:
               version: 2.0.1
             - compatibilityRange: "[2.5.0-M1,3.0.0-M1)"
               version: 2.1.6
-            - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
-              version: 3.0.3
+            - compatibilityRange: "[3.0.0-M1,3.1.0)"
```
snicoll, line 444: What is this supposed to do? Do you mean to include this version in the `3.1.x` line? If so, this is the wrong way to do it.

Your change means that the starter would be available for all milestones of `3.1.0`, excluding `3.1.0`. Concretely, this wouldn't have any effect as those milestones are not available as an option now it has GA'ed. 

If you meant to include  the starter for Spring Boot `3.1.x`, please refer to the other examples, it should be `3.2.0-M1)` meaning up to the first milestone of 3.2.x excluded.

Let me know what the intention was and I can polish, or you can push an additional commit.

arvindkrishnakumar-okta: Intention is to include the starter for Spring Boot 3.1.x and I've updated the PR. Thanks!

PR #1205, opened 2023-05-31
Upgrade to Testcontainers 1.18.3

eddumelendez:
https://repo1.maven.org/maven2/org/testcontainers/testcontainers/1.18.3/testcontainers-1.18.3.pom


eddumelendez:
As of today, most spring boot integrations provide testcontainers dependency already via the Initializr. I was thinking about stopping the upgrade once the Spring Boot 3.0.x is EOL but if it is decided to freeze the version early then it totally make sense to me. Everything new in Testcontainers should be in versions >= Spring Boot 3.1.x

BTW, this is an important one due to a bug fix released in the previous version.

PR #1215, opened 2023-06-06
Upgrade to Spring Cloud GCP 4.4.0 and 3.5.2

mpeddada1:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -109,9 +109,9 @@ initializr:
           - compatibilityRange: "[2.4.0-M1,2.6.0-M1)"
             version: 2.0.11
           - compatibilityRange: "[2.6.0-M1,3.0.0-M1)"
-            version: 3.5.1
+            version: 3.5.2
           - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
-            version: 4.3.1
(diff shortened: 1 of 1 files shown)


snicoll:
@mpeddada1 looking at the [release notes for `4.4.0`](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/releases/tag/v4.4.0), it looks like this version is compatible with Spring Boot 3.1 (given the upgrade to Spring Cloud). Shouldn't we allow 4.4.0 to be used with Spring Boot 3.1 then? Currently, it is disabled.

snicoll:
I made a judgment call based on https://github.com/GoogleCloudPlatform/spring-cloud-gcp#compatibility-with-spring-project-versions and enabled Spring Cloud GCP 4.4.0 on Spring Boot 3.1. @elefeint if that looks wrong please let me know and we can adjust.

elefeint:
@meltsufin? Sorry, I left the project, so I don't actually know what the right thing here is.

meltsufin:
@snicoll We are still trying to figure out if we can support Spring Boot 3.0 and 3.1 in the same branch. The [PR to update to 3.1](https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/1874) is currently red.

PR #1224, opened 2023-06-21
Add support for Hilla

platosha:
Hello! Let me kindly ask for adding the Hilla entry again, now that we have added support for Gradle in the Hilla 2.1 feature release.

This PR is a follow-up for #1113, which was reverted previously. Also see the related entry proposal: #872

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -53,6 +53,13 @@ initializr:
             version: 2.7.4
           - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
             version: 3.0.4
+      hilla:
+        groupId: dev.hilla
+        artifactId: hilla-bom
+        versionProperty: hilla.version
+        mappings:
+          - compatibilityRange: "[3.1.0-M1,3.2.0-M1)"
+            version: 2.1.1
       solace-spring-boot:
         groupId: com.solace.spring.boot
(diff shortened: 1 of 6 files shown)


start-client/dev/api.mock.json
```diff
@@ -176,6 +176,18 @@
               },
               "reference": { "href": "https://vaadin.com/docs" }
             }
+          },
```
snicoll, line 179: This is a mock file for testing and doesn't need to be updated.

start-site/src/main/resources/application.yml
```diff
@@ -163,6 +163,13 @@ initializr:
             version: 24.0.7
           - compatibilityRange: "[3.1.0-M1,3.2.0-M1)"
             version: 24.1.0
+      hilla:
```
snicoll, line 166: BOMs are ordered alphabetically.

snicoll (commented):
Thanks for the PR. I would need to play with this a bit more but I've added a few comments for your consideration.

platosha:
> Thanks for the PR. I would need to play with this a bit more but I've added a few comments for your consideration.

Thank you. Addressed the comments so far.

snicoll:
@platosha can you please squash all that and rebase on top of `main` please?

platosha:
@snicoll sure, done so

PR #1225, opened 2023-06-21
Support MyBatis testing starter 

kazuki43zoo:
The [mybatis-spring-boot-starter](http://mybatis.org/spring-boot-starter/mybatis-spring-boot-autoconfigure/) has been supported on `start.spring.io`.
Adding, I propose to add the MyBatis testing support starter([mybatis-spring-boot-starter-test](http://mybatis.org/spring-boot-starter/mybatis-spring-boot-test-autoconfigure/)) for quick start with MyBatis slice testing.  It provide the `@MybatisTest` annotation for slice testing, version support matrix is same with the mybatis-spring-boot-starter.

WDYT?

If need to create new `entry-proposal` issue, I will create it.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/DependencyProjectGenerationConfiguration.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/DependencyProjectGenerationConfiguration.java
@@ -26,6 +26,7 @@ import io.spring.initializr.generator.spring.build.gradle.ConditionalOnGradleVer
 import io.spring.initializr.metadata.InitializrMetadata;
 import io.spring.start.site.extension.dependency.liquibase.LiquibaseProjectContributor;
 import io.spring.start.site.extension.dependency.lombok.LombokGradleBuildCustomizer;
+import io.spring.start.site.extension.dependency.mybatis.MyBatisTestBuildCustomizer;
 import io.spring.start.site.extension.dependency.okta.OktaHelpDocumentCustomizer;
 import io.spring.start.site.extension.dependency.reactor.ReactorTestBuildCustomizer;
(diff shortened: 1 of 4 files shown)


wilkinsona:
Rather than this being a new dependency that a user has to select, I wonder if we should automatically add a dependency on  `mybatis-spring-boot-starter-test` when the user selects the existing MyBatis entry?

@kazuki43zoo Is there any reason why someone would not want a test-scoped dependency on `mybatis-spring-boot-starter-test` when they're using `mybatis-spring-boot-starter`?

kazuki43zoo:
> Rather than this being a new dependency that a user has to select, I wonder if we should automatically add a dependency on `mybatis-spring-boot-starter-test` when the user selects the existing MyBatis entry?
> 

I think so!!

> @kazuki43zoo Is there any reason why someone would not want a test-scoped dependency on `mybatis-spring-boot-starter-test` when they're using `mybatis-spring-boot-starter`?

Nothing!! I didn't know how to add together mybatis-spring-boot-starter and mybatis-spring-boot-starter-test. If can adding together, it is best solution!!

wilkinsona:
It can be done with a `BuildCustomizer`. For example, [here's one](https://github.com/spring-io/start.spring.io/blob/9d412386bab254441d81246dfeee7d404362f90c/start-site/src/main/java/io/spring/start/site/extension/dependency/springsecurity/SpringSecurityTestBuildCustomizer.java) that adds a dependency on Spring Security's test module when Spring Security has been selected. Would you like to update your PR to take a similar approach?

kazuki43zoo:
@wilkinsona Thanks for your advice!! I will try updating this PR!!

kazuki43zoo:
@wilkinsona I updated this PR.

snicoll:
@kazuki43zoo very nice thanks!

PR #1228, opened 2023-06-23
Add Spring Modulith

odrotbohm:
This commit adds support to add Spring Modulith to a project. We add the Event Publication Registry matching the (optionally) also selected persistence flavor (JPA, JDBC, MongoDB). Furthermore, the observability support in case any of the supported observability backends is selected as dependency.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
@@ -1,0 +1,104 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.spring.start.site.extension.dependency.springmodulith;
+
+import java.util.Collection;
+import java.util.List;
+
+import io.spring.initializr.generator.buildsystem.Build;
+import io.spring.initializr.generator.buildsystem.Dependency;
+import io.spring.initializr.generator.buildsystem.Dependency.Builder;
+import io.spring.initializr.generator.buildsystem.DependencyContainer;
+import io.spring.initializr.generator.buildsystem.DependencyScope;
+import io.spring.initializr.generator.spring.build.BuildCustomizer;
+
+/**
+ * Registers Spring Modulith dependencies to the build file of the project to be created,
+ * in particular registering integration dependencies depending on others registered for
+ * inclusion as well (observability and persistence).
+ *
+ * @author Oliver Drotbohm
+ */
+class SpringModulithBuildCustomizer implements BuildCustomizer<Build> {
+
+	static final String GROUP_ID = "org.springframework.experimental";
+	static final String BOM_ARTIFACT_ID = "spring-modulith-bom";
+
+	static final Collection<String> OBSERVABILITY_DEPENDENCIES = List.of("datadog", "graphite", "influx", "new-relic",
+			"prometheus", "wavefront", "zipkin");
+
+	private static final String ARTIFACT_PREFIX = "spring-modulith-";
+
+	private static final Builder<?> ACTUATOR = Dependency.withCoordinates(GROUP_ID, ARTIFACT_PREFIX + "actuator")
+		.scope(DependencyScope.RUNTIME);
+
+	private static final Builder<?> OBSERVABILITY = Dependency
+		.withCoordinates(GROUP_ID, ARTIFACT_PREFIX + "observability")
+		.scope(DependencyScope.RUNTIME);
+
+	private static final Builder<?> STARTER_JDBC = Dependency.withCoordinates(GROUP_ID,
+			ARTIFACT_PREFIX + "starter-jdbc");
+
+	private static final Builder<?> STARTER_JPA = Dependency.withCoordinates(GROUP_ID, ARTIFACT_PREFIX + "starter-jpa");
+
+	private static final Builder<?> STARTER_MONGODB = Dependency.withCoordinates(GROUP_ID,
+			ARTIFACT_PREFIX + "starter-mongodb");
+
+	private static final Builder<?> TEST = Dependency.withCoordinates(GROUP_ID, ARTIFACT_PREFIX + "starter-test")
+		.scope(DependencyScope.TEST_COMPILE);
+
+	@Override
+	public void customize(Build build) {
+		DependencyContainer dependencies = build.dependencies();
+
+		// Actuator
+
+		if (dependencies.has("actuator")) {
+			dependencies.add("modulith-actuator", ACTUATOR);
+		}
+
+		// Observability
+
+		if (OBSERVABILITY_DEPENDENCIES.stream().anyMatch(dependencies::has)) {
+			dependencies.add("modulith-observability", OBSERVABILITY);
+		}
+
+		// Event publication registry support
+
+		addEventPublicationRegistryBackend(build);
+
+		dependencies.add("modulith-starter-test", TEST);
+	}
+
+	private void addEventPublicationRegistryBackend(Build build) {
+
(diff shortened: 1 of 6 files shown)


start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
```diff
@@ -0,0 +1,110 @@
...
+
+	private static final Builder<?> STARTER_MONGODB = Dependency.withCoordinates(GROUP_ID,
+			ARTIFACT_PREFIX + "starter-mongodb");
+
+	private static final Builder<?> TEST = Dependency.withCoordinates(GROUP_ID, ARTIFACT_PREFIX + "starter-test")
+			.scope(DependencyScope.TEST_COMPILE);
+
+	@Override
+	public void customize(Build build) {
+
```
wilkinsona, line 60: This projects's style is to not have a blank line at the start of a method.

start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
```diff
@@ -0,0 +1,110 @@
...
+		}
+
+		// Event publication registry support
+
+		var added = addEventPublicationRegistryBackend(build);
+
+		if (added) {
+			dependencies.remove("modulith");
+			build.boms().add("spring-modulith");
+		}
```
wilkinsona, line 82: I wonder if we should just keep the core starter dependency rather than swapping them.

wilkinsona (changes requested):

start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
odrotbohm: I originally wanted to avoid declaring dependencies that do not explicitly enable features, but the core starter actually enables the Moments (event-based abstraction of time) feature. With that thought in mind, it's reasonable to still explicitly include it. I've reworked the section accordingly.

start-site/src/main/resources/application.yml
```diff
@@ -139,6 +139,14 @@ initializr:
             version: 3.5.0
           - compatibilityRange: "[3.0.0,3.1.0-M1)"
             version: 4.0.0
+      spring-modulith:
+        groupId: org.springframework.modulith
+        artifactId: spring-modulith-bom
+        version: 1.0.0-M1
```
snicoll, line 145: You don't need a version here if you provide it through a mapping.

start-site/src/main/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizer.java
```diff
@@ -0,0 +1,101 @@
...
+		// Event publication registry support
+
+		addEventPublicationRegistryBackend(build);
+
+		dependencies.add("modulith-starter-test", TEST);
+	}
+
+	private void addEventPublicationRegistryBackend(Build build) {
+
+		var dependencies = build.dependencies();
```
snicoll, line 87: There was a previous review about the use of `var`. Can you please review the  code to remove any remaining use?

start-site/src/main/resources/application.yml
```diff
@@ -199,12 +207,26 @@ initializr:
       spring-releases:
         name: Spring Releases
         url: https://repo.spring.io/release
+      spring-milestones:
```
snicoll, line 210: You don't need this. `spring-milestones` and `spring-snapshots` are automatically available.

start-site/src/main/resources/application.yml
```diff
@@ -199,12 +207,26 @@ initializr:
...
+        name: Spring Milestones
+        url: https://repo.spring.io/milestone
       sonatype-snapshots:
         name: Sonatype Snapshots
         url: https://oss.sonatype.org/content/repositories/snapshots/
         releasesEnabled: false
         snapshotsEnabled: true
   dependencies:
+    - name: Architecture
```
snicoll, line 219: I'd not create a new group for a single entry for now. Can you please move it to the Developer Tools group?

start-site/src/main/resources/application.yml
```diff
@@ -199,12 +207,26 @@ initializr:
...
+      content:
+        - name: Spring Modulith
```
snicoll, line 221: We don't name things with "Spring" in front of it. Look at other entries for more details.

start-site/src/test/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizerTests.java
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * @author Oliver Drotbohm
+ */
+class SpringModulithBuildCustomizerTests extends AbstractExtensionTests {
+
+	private final SpringModulithBuildCustomizer customizer = new SpringModulithBuildCustomizer();
+
+	@Test
+	void registersCoreStarterByDefault() {
+
```
snicoll, line 48: We don't add new line there (same remark as the use of `var`).

start-site/src/test/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizerTests.java
```diff
@@ -0,0 +1,117 @@
...
+		Build build = createBuild();
+		build.dependencies().add("actuator");
+
+		customizer.customize(build);
+
+		assertThat(build.dependencies().ids()).contains("modulith-actuator");
+	}
+
+	@TestFactory
+	Stream<DynamicTest> registersObservabilityStarterIfObservabilityDependencyIsPresent() {
```
snicoll, line 76: This is the first use of this in the project. Rather than using a new testing style, I'd like to make sure it can't be achieved with the techniques we're using at the moment.

snicoll (changes requested):
Thanks for the PR. I've added a few comments for your consideration.

I know building the client is an issue due to installing `node-gyp`. If you don't want to install it, you can remove the following dependency locally:

```xml
<dependency>
  <groupId>io.spring.start</groupId>
  <artifactId>start-client</artifactId>
  <scope>runtime</scope>
</dependency>
```

And then run the following command:  `./mvnw verify -pl start-site`. Thanks!

start-site/src/main/resources/application.yml
odrotbohm: Are you sure about that? I see all Spring Data flavors using the prefix, Spring Integration, Spring Batch etc.

snicoll: My mistake, this was my recollection before we did some harmonization. Please disregard this.

start-site/src/test/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizerTests.java
odrotbohm: Are you saying you prefer 7 copy & pasted test cases that only differ in the input string? Is `@ParameterizedTest` something that'd be acceptable?

start-site/src/main/resources/application.yml
odrotbohm: All good, no worries.

start-site/src/test/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizerTests.java
odrotbohm: Apparently, that is, as I've used it for the test below. Will refactor the test accordingly.

odrotbohm:
That's been a useful tip, thanks! I have incorporated your feedback, fixed the Checkstyle issues and updated the branch.

start-site/src/test/java/io/spring/start/site/extension/dependency/springmodulith/SpringModulithBuildCustomizerTests.java
snicoll: > Are you saying you prefer 7 copy & pasted test cases that only differ in the input string

No. I was talking about `DynamicTest`.

snicoll:
Unfortunately, the build is failing for me and I can still see a bunch of additional newlines in method bodies. Can you please look again?

odrotbohm:
Sorry, for the hassle. Done.

snicoll:
I've taken over now so there's no need to push further updates. I've created https://github.com/spring-projects/spring-modulith/issues/237 as the core starter doesn't bring the "core" spring boot starter, which breaks the metadata test when this PR is around.

I've also found some lingering `experimental` in the build customizer. None of those are a blocker.

PR #1237, opened 2023-07-04
Upgrade to Vaadin 24.1.2, 24.0.9, 23.3.16, 14.10.3

ZheSun88:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -161,15 +161,15 @@ initializr:
         versionProperty: vaadin.version
         mappings:
           - compatibilityRange: "[2.1.0.RELEASE,2.6.0-M1)"
-            version: 14.10.2
+            version: 14.10.3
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -161,15 +161,15 @@ initializr:
         versionProperty: vaadin.version
         mappings:
           - compatibilityRange: "[2.1.0.RELEASE,2.6.0-M1)"
-            version: 14.10.2
+            version: 14.10.3
           - compatibilityRange: "[2.6.0-M1,2.7.0-M1)"
-            version: 23.2.15
+            version: 23.2.14
```
wilkinsona, line 166: This looks like a downgrade. Should it be `.16`?

ZheSun88: oops, that is a mistake.. actually 23.2 is not supported any more. i should not touch that version.

PR #1238, opened 2023-07-04
Upgrade to Camel 4.0.0-RC1

davsclaus:
https://camel.apache.org/blog/2023/07/RELEASE-4.0.0-RC1/

This version is upgraded to Spring Boot 3.1.1

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -867,6 +867,8 @@ initializr:
               version: 3.20.6
             - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
               version: 4.0.0-M3
+            - compatibilityRange: "[3.1.0,3.2.0-M1)"
+              version: 4.0.0-RC1
           description: Apache Camel is an open source integration framework that empowers you to quickly and easily integrate various systems consuming or producing data.
           groupId: org.apache.camel.springboot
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -867,6 +867,8 @@ initializr:
               version: 3.20.6
             - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
               version: 4.0.0-M3
+            - compatibilityRange: "[3.1.0,3.2.0-M1)"
```
wilkinsona, line 870: To avoid a gap in the version ranges, this should use 3.1.0-M1 as the lower bound.

start-site/src/main/resources/application.yml
```diff
@@ -867,6 +867,8 @@ initializr:
               version: 3.20.6
             - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
               version: 4.0.0-M3
```
wilkinsona, line 869: Should this change too or is it intentional that Spring Boot 3.0.x users continue to use Camel 4.0.0-M3?

davsclaus: Ah yeah those Milestone releases are essentially no longer supported, and ideally we would only have the latest 4.0-XX until we have a GA release.

I can do a new PR

PR #1239, opened 2023-07-05
Upgrade to Camel 4.0.0-RC1

davsclaus:
https://camel.apache.org/blog/2023/07/RELEASE-4.0.0-RC1/

This version is upgraded to Spring Boot 3.1.1

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -865,8 +865,8 @@ initializr:
               version: 3.17.0
             - compatibilityRange: "[2.7.0.M1,3.0.0-M1)"
               version: 3.20.6
-            - compatibilityRange: "[3.0.0-M1,3.1.0-M1)"
-              version: 4.0.0-M3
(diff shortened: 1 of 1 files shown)


davsclaus:
PS: Do the older version of spring boot still exists on start.spring.io, or can we "cleanup" and remove old Camel versions in the list?

wilkinsona:
I think we could do something like https://github.com/spring-io/start.spring.io/commit/f688a09992a35469e4ac56fbc79537a2ebd4acaa, making 2.7 the minimum version. Before doing anything, I'd like to double-check with @snicoll when he's back in the office.

snicoll:
I like to keep these around as start.spring.io/info is a good tool to know which versions were used for folks doing intermediate updates. However, the lowest bound is `2.3.0`, I think we can at least bump to 2.5.0, or even 2.6. I've created https://github.com/spring-io/start.spring.io/issues/1241

PR #1262, opened 2023-07-31
Add Testcontainers Solace module when solace and testcontainers are selected

eddumelendez:
Testcontainers offers a solace module that can be used when
both technologies are selected.


snicoll:
Ahem. I've just realized the auto-configuration on the other side needs to handle `@ServiceConnection`. So we're a bit stuck, aren't we? I'll create an issue to see if there's interest.

snicoll:
I've created https://github.com/SolaceProducts/solace-spring-boot/issues/106

eddumelendez:
Looking forward to see the ServiceConnection support in Solace Spring Boot so it can be autogenerated in start.spring.io. Thanks for raising the issue!

PR #1264, opened 2023-07-31
Add Testcontainers GCloud module when GCP messaging and testcontainers are selected

eddumelendez:
Testcontainers offers a GCloud module where `PubSubEmulatorContainer`
can be found and be used for testing.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/testcontainers/TestcontainersModuleRegistry.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/testcontainers/TestcontainersModuleRegistry.java
@@ -41,6 +41,8 @@ abstract class TestcontainersModuleRegistry {
 		return create(
 				onDependencies("amqp").customizeBuild(addModule("rabbitmq"))
 					.customizeHelpDocument(addReferenceLink("RabbitMQ Module", "rabbitmq/")),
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks again @eddumelendez. Looking at the website the module states:

> Testcontainers module for the Google Cloud Platform's [Cloud SDK](https://cloud.google.com/sdk/).

Is there a reason why the PR constrains it to only gcp pub sub? At the very least, GCP support should also include it as it is a segway to configure the project with GCP.

eddumelendez:
> Is there a reason why the PR constrains it to only gcp pub sub? At the very least, GCP support should also include it as it is a segway to configure the project with GCP.

Yeah, make sense now that I think about it. I tried to be specific because GCP Messaging is the one that explicitly can provide the Testcontainers class. However, if as user I know I'll use Datastore, Firestore or Spanner I can select GCP support and Testcontainers which will provide the dependency too and later I can add the proper starter.

I can update the PR later today.

eddumelendez:
PR updated. Now, if cloud-gcp is selected, testcontainers gcloud module is also added.

BTW, there is already an issue in spring-cloud-gcp to support ServiceConnections. Seehttps://github.com/GoogleCloudPlatform/spring-cloud-gcp/issues/1780

snicoll:
Thanks for the update @eddumelendez. I've polished the proposal to avoid adding the help link twice if both gcp and gcp-pub-sub are selected.

PR #1265, opened 2023-07-31
Add flyway-database-oracle dependency with spring boot 3.2.0-M1

eddumelendez:
Flyway 9.19.0 added a new depedency for oracle and Spring Boot 3.2.0-M1
started using Flyway 9.20.1. Initializr will add this new dependency
when oracle and flyway are selected along with Spring Boot 3.2.0-M1


snicoll:
Good catch, thanks again @eddumelendez!

PR #1285, opened 2023-08-21
Add io.asyncer:r2dbc-mysql when spring boot 3.1 is selected

eddumelendez:
`io.asyncer:r2dbc-mysql` was added in the dependency management
in spring boot 3.1. Currently, when r2dbc and mysql are selected
the generated project doesn't include the new dependepency. This
commit fix the behavior.


snicoll:
Thanks again @eddumelendez

PR #1291, opened 2023-09-06
Add support for Pulsar container

eddumelendez:
Recently, spring pulsar was integrated in spring boot with
`ConnectionDetails` and `ServiceConnection`.

PR #1294, opened 2023-09-08
Add optional 'command' to Docker compose service

onobc:
Depends on https://github.com/spring-io/initializr/pull/1461

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/container/DockerService.java
+++ b/start-site/src/main/java/io/spring/start/site/container/DockerService.java
@@ -24,6 +24,7 @@ import io.spring.initializr.generator.container.docker.compose.ComposeService.Bu
  * Description of a Docker service.
  *
  * @author Stephane Nicoll
+ * @author Chris Bono
  */
 public class DockerService implements Consumer<Builder> {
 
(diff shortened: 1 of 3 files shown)


snicoll:
@onobc I am totally cool with adding a command option but both PRs don't have a description. What's the need here for this?

wilkinsona:
Chris and I discussed this a bit yesterday. It's for Pulsar which requires a custom command be used to launch it in standalone mode. It's to enable the equivalent of [this code in Testcontainers' `PulsarContainer`](https://github.com/testcontainers/testcontainers-java/blob/a7737fdee46dec6bc2441faaff009bd17480efed/modules/pulsar/src/main/java/org/testcontainers/containers/PulsarContainer.java#L97-L102).

start-site/src/main/java/io/spring/start/site/container/DockerService.java
```diff
@@ -33,12 +34,15 @@ public class DockerService implements Consumer<Builder> {
 
 	private final String website;
 
+	private final String command;
+
 	private final int[] ports;
 
-	DockerService(String image, String imageTag, String website, int... ports) {
+	DockerService(String image, String imageTag, String website, String command, int... ports) {
```
wilkinsona, line 41: Overload this instead so that `SimpleDockerServiceResolver` doesn't have to pass in `null` each time?

snicoll: Yeah, I was thinking the same. That or we should maybe introduce a builder if we keep getting optional attributes there.

snicoll:
Thanks @wilkinsona. I guess this should answer this question then https://github.com/spring-io/start.spring.io/pull/1291#issuecomment-1711359136

start-site/src/main/java/io/spring/start/site/container/DockerService.java
onobc: Adding a builder.

PR #1322, opened 2023-10-21
Add Pulsar Binder to Spring Cloud Stream

onobc:
This commit ensures the Spring Cloud Stream Pulsar binder dependency is automatically added when the user chooses Spring Boot >= 3.2.0-M3 and adds the Pulsar and Cloud Stream dependencies.

This was not an issue in the older Spring Pulsar `0.2.x` offering because the binder was contained in the Spring Pulsar repo. Now the binder has been moved into SCSt proper, the binder dependency must be added as well. I missed this detail in my original update to Pulsar/Boot32.

### DIFF
--- a/start-site/src/test/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudStreamBuildCustomizerTests.java
+++ b/start-site/src/test/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudStreamBuildCustomizerTests.java
@@ -34,6 +34,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  *
  * @author Stephane Nicoll
  * @author Brian Clozel
+ * @author Chris Bono
  */
 class SpringCloudStreamBuildCustomizerTests extends AbstractExtensionTests {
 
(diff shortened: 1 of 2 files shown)


start-site/src/test/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudStreamBuildCustomizerTests.java
```diff
@@ -85,6 +89,19 @@ void springCloudStreamWithKafkaStreams(Version springBootVersion, Dependency tes
 			.hasDependenciesSize(5);
 	}
 
+	@ParameterizedTest
+	@MethodSource("springCloudStreamWithPulsarArguments")
```
onobc, line 93: I am not super excited of screaming "...WithPulsar" here but the variations we are now trying to test are:
- Boot 2.7.x (the test binder deps differ across 2.x and 3.x)
- Boot 3.x (test test binder deps differ across 2.x and 3.x)
- Boot 3.2.0-M3++ (new binder was added here)

Because the Pulsar binder does not exist <=3.2.0-M3 we can simply add more args to the test params for boot32x as the behavior is additive (iow new binder dep).

snicoll: I am not a big fan of those either. I've polished to make it short for now. We can revisit those tests once we remove support for 2.x which is soon.

PR #1348, opened 2023-11-14
Add Netflix DGS Codegen plugin support

bclozel:
This commit adds a new dependency to start.spring.io, the "Netflix Data
Graph Service Codegen". This will not add a dependency to generated
project, but rather configure a Maven or a Gradle plugin in the build.

This Codegen plugin parses GraphQL schema files describing remote
GraphQL APIs and generates Java classes to easily create queries and
deserialize GraphQL responses.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.spring.start.site.extension.dependency.dgs;
+
+import io.spring.initializr.generator.buildsystem.maven.MavenBuild;
+import io.spring.initializr.generator.spring.build.BuildCustomizer;
+
+/**
+ * {@link BuildCustomizer} for Maven projects using DGS Codegen (community maintained).
+ *
+ * @author Brian Clozel
+ * @see <a href="https://github.com/deweyjose/graphqlcodegen">Maven GraphQL Codegen
+ * plugin</a>
+ */
+public class DgsCodegenMavenBuildCustomizer implements BuildCustomizer<MavenBuild> {
+
+	private final String pluginVersion;
+
+	private final String packageName;
+
+	public DgsCodegenMavenBuildCustomizer(String pluginVersion, String packageName) {
+		this.pluginVersion = pluginVersion;
+		this.packageName = packageName;
+	}
+
+	@Override
+	public void customize(MavenBuild build) {
+		build.plugins()
+			.add("io.github.deweyjose", "graphqlcodegen-maven-plugin",
+					(plugin) -> plugin.version(this.pluginVersion)
+						.execution("dgs-codegen",
+								(execution) -> execution.goal("generate")
+									.configuration((configuration) -> configuration
+										.add("schemaPaths",
+												(schemaPaths) -> schemaPaths.add("param",
+														"src/main/resources/graphql-client"))
+										.add("packageName", this.packageName + ".codegen")
+										.add("addGeneratedAnnotation", "true"))));
+		build.plugins()
+			.add("org.codehaus.mojo", "build-helper-maven-plugin",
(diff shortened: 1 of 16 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -203,6 +203,16 @@ initializr:
           artifactId: spring-boot
           description: Support for compiling Spring applications to native executables using the GraalVM native-image compiler.
           starter: false
+        - name: GraphQL DGS Code Generation
+          id: dgs-codegen
+          compatibilityRange: "3.0.0-M1"
+          groupId: com.netflix.graphql.dgs.codegen
+          artifactId: graphql-dgs-codegen-gradle
+          mappings:
```
snicoll, line 211: Wondering if we need a mapping here considering that the artifact is just a trigger for the build system setup.

start-site/src/test/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenHelpDocumentCustomizerTests.java
```diff
@@ -0,0 +1,66 @@
...
+class DgsCodegenHelpDocumentCustomizerTests extends AbstractExtensionTests {
+
+	@Autowired
+	private MustacheTemplateRenderer templateRenderer;
+
+	@Test
+	void mavenBuildAddsLinkToMavenCodegenPlugin() {
+
+		assertHelpDocument("maven-project", "dgs-codegen").contains("## GraphQL code generation with DGS")
+			.containsSubsequence("This project has been configured to use the Netflix DGS Codegen plugin.",
```
snicoll, line 42: I wonder if that's strictly necessary as any wording change needs the test to be updated. I think we're probably a bit more relaxed for other help document tests. Checking that the proper link according to the build system is specified (and the non matching one is not present) looks the real important bit for me.

start-site/src/test/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenProjectGenerationConfigurationTests.java
```diff
@@ -0,0 +1,143 @@
...
+
+/**
+ * Tests for {@link DgsCodegenProjectGenerationConfiguration}.
+ *
+ * @author Brian Clozel
+ */
+class DgsCodegenProjectGenerationConfigurationTests extends AbstractExtensionTests {
+
+	@Test
+	void gradleBuildWithoutNativeDoesNotConfigureNativeBuildTools() {
```
snicoll, line 33: Typo as it refers to native?

start-site/src/test/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenProjectGenerationConfigurationTests.java
```diff
@@ -0,0 +1,143 @@
...
+		ProjectRequest request = createProjectRequest("web");
+		assertThat(gradleBuild(request)).doesNotContain("com.netflix.dgs.codegen");
+	}
+
+	@Test
+	void mavenBuildWithoutNativeDoesNotConfigureNativeBuildTools() {
```
snicoll, line 39: Typo as it refers to native?

start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
```diff
@@ -0,0 +1,62 @@
...
+									.configuration((configuration) -> configuration
+										.add("schemaPaths",
+												(schemaPaths) -> schemaPaths.add("param",
+														"src/main/resources/graphql-client"))
+										.add("packageName", this.packageName + ".codegen")
+										.add("generateDataTypes", "false")
+										.add("addGeneratedAnnotation", "true"))));
+		build.plugins()
+			.add("org.codehaus.mojo", "build-helper-maven-plugin",
+					(plugin) -> plugin.execution("generate-sources", (execution) -> execution.goal("add-source")
```
snicoll, line 56: I wonder if `generate-sources` is a suitable `id` for this. I think `add-dgs-source` or something could be better?

snicoll (commented):
this look complete and quite polished already. I've just added a few nits for your consideration. Thanks!

start-site/src/main/resources/application.yml
bclozel: Right now the only added benefit is that if we upgrade the plugin version and it's not fully published on Maven Central, our verification tests will catch it. I'd be tempted to leave that in for this reason.

start-site/src/test/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenHelpDocumentCustomizerTests.java
bclozel: I've relaxed the tests, thanks!

start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
bclozel: Changed the code generation plugin configuration as well.

bclozel:
This PR is still in draft mode, we're discussing some details here and maybe we should link to Spring for GraphQL's docs once https://github.com/spring-projects/spring-graphql/issues/847 is resolved.

PR #1363, opened 2023-11-28
Upgrade to Vaadin 24.2.5 and hilla 2.4.0

manolo:


snicoll:
Hey @manolo, thanks for that. Spring Boot 3.2 support is due for Vaadin 24.3.x and Hila 2.5.x, correct?

PR #1373, opened 2023-12-13
Upgrade to Spring Cloud GCP 4.9.0

mpeddada1:


snicoll:
@mpeddada1 and @meltsufin it looks like this version is Spring Boot 3.2 compatible, https://github.com/GoogleCloudPlatform/spring-cloud-gcp/pull/2397. If that is the case, we need to know to open up the range.

mpeddada1:
@snicoll Thanks for the ping! This version is not yet fully compatible with v3.2.0 so leaving the range as-is for now.

PR #1376, opened 2023-12-14
Move Pulsar container to 'latest' tag

onobc:
This commit updates the Pulsar Docker container version from `3.1.0` to `3.1.1` for both the Docker Compose file and the Testcontainers generated container definition.

For reference, Spring Boot uses [3.1.1](https://github.com/spring-projects/spring-boot/blob/b62b6d56c180cd0fba140fc7a73bfe2d691bea6c/spring-boot-project/spring-boot-dependencies/build.gradle#L1140) for its recommended version.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/container/SimpleDockerServiceResolver.java
+++ b/start-site/src/main/java/io/spring/start/site/container/SimpleDockerServiceResolver.java
@@ -113,8 +113,7 @@ public class SimpleDockerServiceResolver implements DockerServiceResolver {
 	}
 
 	private static DockerService pulsar() {
-		// The latest tag they provide is not the 'latest' GA
-		return DockerService.withImageAndTag("apachepulsar/pulsar:3.1.1")
+		return DockerService.withImageAndTag("apachepulsar/pulsar")
 			.website("https://hub.docker.com/r/apachepulsar/pulsar")
 			.command("bin/pulsar standalone")
 			.ports(8080, 6650)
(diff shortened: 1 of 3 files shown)


snicoll:
If there is a relationship between the container version and the library version, we should not hardcode the version here but rather read the version from the BOM. 

Is there any reason we're not doing that?

onobc:
>If there is a relationship between the container version and the library version, we should not hardcode the version here but rather read the version from the BOM.

> Is there any reason we're not doing that?

Good point. There is container created that corresponds to the version but I believe it is not released at the same time as the library. I have reached out to the Pulsar team for more details on this process and what guarantees there are (if any). Would it be worth adding a managed version in the BOM for the pulsar container? What are the other supported techs doing in this area (or do they all have a 1:1 relationship between library and container versions)?

onobc:
@snicoll I got feedback from the Pulsar team, w/ the latest release process, the containers are released with the library, but they show up on Docker hub sometime after the library is available on Maven central. I think for our use we can rely on this mapping since by the time we pull the updated library version, the container is likely available (or will be shortly). If you agree, I will remove the hardcoding in this PR and instead get the version from the SB recommended version.

snicoll:
Sounds good, thanks!

onobc:
@snicoll digging a bit further, the Pulsar containers are now properly using 'latest' tag to match the actual "latest" version.  Reduced the complexity here accordingly.

PR #1390, opened 2024-01-10
Upgrade to Timefold Solver 1.6.0

triceo:
Introducing new release of Timefold Solver for the website.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -139,7 +139,7 @@ initializr:
         versionProperty: timefold-solver.version
         mappings:
           - compatibilityRange: "[3.0.0,3.3.0-M1)"
-            version: 1.5.0
+            version: 1.6.0
       vaadin:
         groupId: com.vaadin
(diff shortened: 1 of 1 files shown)


triceo:
@snicoll Thanks for merging!
Quick question - we would like these PRs to be generated automatically by our CI, but I'm struggling with one aspect of it.

If it's generated automatically, it will be generated by our bot account. The account certainly doesn't have a CLA signed. (Can a bot even sign a CLA?) What should we do in this situation?

Is filing an issue a better approach?

snicoll:
Automation for that sort of thing can be tricky. Does the automation wait for the bits to be available on Maven Central? (this is a requirement). Creating an issue right behind your release without making sure the bits can be consumer by central is also problematic. 

Changes to version numbers are considered trivial changes so it doesn't technically require the CLA.

triceo:
Yes, the PR is only triggered when the bits are on Maven Central.
If it's OK that the PR comes from a bot account without CLA, I will make it happen.

PR #1392, opened 2024-01-16
Upgrade to Spring Boot Admin 3.2.1

SteKoe:
We, of course, support Spring Boot 3.2 as well :)

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -43,8 +43,8 @@ initializr:
         artifactId: spring-boot-admin-dependencies
         versionProperty: spring-boot-admin.version
         mappings:
-          - compatibilityRange: "[3.1.0,3.2.0-M1)"
-            version: 3.1.5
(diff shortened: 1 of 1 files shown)


start-site/src/main/resources/application.yml
```diff
@@ -936,7 +936,7 @@ initializr:
           id: sentry
           bom: sentry
           description: Application performance monitoring and error tracking that help software teams see clearer, solve quicker, and learn continuously.
-          compatibilityRange: "[3.0.0,3.2.0-M1)"
+          compatibilityRange: "[3.0.0,3.3.0-M1)"
```
snicoll, line 939: That looks wrong to me.

snicoll (changes requested):
Thanks but I think you've changed an unrelated entry.

start-site/src/main/resources/application.yml
SteKoe: My bad, sorry! I have reverted the change.

PR #1415, opened 2024-02-21
Add ActiveMQ module when ActiveMQ/Artemis and Testcontainers are selected

eddumelendez:
Testcontainers 1.19.4 offers an ActiveMQ module, which provides two implementations
`ActiveMQContainer` and `ArtemisContainer`. Both, are already supported as part of
Spring Boot 3.3.0-M2.


start-site/src/main/java/io/spring/start/site/container/SimpleDockerServiceResolver.java
```diff
@@ -57,6 +60,20 @@ private static DockerService activeMQ() {
...
+	private static DockerService activeMQClassic() {
+		return DockerService.withImageAndTag("apache/activemq-classic")
+			.website("https://hub.docker.com/r/apache/activemq-classic")
+			.ports(61616)
+			.build();
+	}
+
+	private static DockerService artemis() {
+		return DockerService.withImageAndTag("apache/activemq-artemis")
+			.website("https://hub.docker.com/r/apache/activemq-classic")
```
eddumelendez, line 72:
```suggestion
			.website("https://hub.docker.com/r/apache/activemq-artemis")
```

start-site/src/main/java/io/spring/start/site/extension/dependency/activemq/ActiveMQProjectGenerationConfiguration.java
```diff
@@ -45,10 +47,26 @@ ServiceConnectionsCustomizer activeMQServiceConnectionsCustomizer(DockerServiceR
 
 	@Bean
 	@ConditionalOnPlatformVersion("3.2.0-M1")
```
mhalbritter, line 49: Shouldn't that read `@ConditionalOnPlatformVersion("3.3.0-M2")`?

eddumelendez: Yes, good catch! I have updated the PR.

mhalbritter:
Thanks! I've tested it, but there are some pieces missing:

When i generate a 3.3.0-SNAPSHOT project, it doesn't include the `activemq` testcontainers module. That's because

```java
private static final VersionRange SPRING_BOOT_3_3_0_OR_LATER = VersionParser.DEFAULT.parseRange("3.3.0");
```

doesn't match `3.3.0-SNAPSHOT`. I guess it should be:

```java
private static final VersionRange SPRING_BOOT_3_3_0_M2_OR_LATER = VersionParser.DEFAULT.parseRange("3.3.0-M2");
```

However, with that change, the generated test main looks like this:

```java
	@Bean
	@ServiceConnection
	ActiveMQContainer<?> activemqContainer() {
		return new ActiveMQContainer<>(DockerImageName.parse("apache/activemq-classic:latest"));
	}
```

When compiling, this fails with:

```java
> Task :compileTestJava FAILED
/home/mhalbritter/Downloads/demo330/src/test/java/com/example/demo330/TestDemo330Application.java:15: error: type ActiveMQContainer does not take parameters
        ActiveMQContainer<?> activemqContainer() {
                         ^
/home/mhalbritter/Downloads/demo330/src/test/java/com/example/demo330/TestDemo330Application.java:16: error: cannot infer type arguments for ActiveMQContainer
                return new ActiveMQContainer<>(DockerImageName.parse("apache/activemq-classic:latest"));
                                            ^
  reason: cannot use '<>' with non-generic class ActiveMQContainer
2 errors

FAILURE: Build failed with an exception.
```

eddumelendez:
Thanks @mhalbritter ! I missed to set the `containerClassNameGeneric` to `false`. Now, it works as expected.

mhalbritter:
Thanks!

PR #1417, opened 2024-02-25
Fix link to Spring Modulith reference.

maciejwalkowiak:
Changed a link to Spring Modulith that appears in `HELP.md` file from https://docs.spring.io/spring-modulith/docs/current/reference/html/ to https://docs.spring.io/spring-modulith/reference/ as the former results in `Page Not Found`.


mhalbritter:
Thanks!

PR #1476, opened 2024-05-24
Upgrade compatibility range for Vaadin

ZheSun88:
we have tested the new Spring-boot 3.3.0 on our product, which can confirm that it is compatible.


mhalbritter:
Thanks!

PR #1490, opened 2024-05-28
Optimizations in Spring Modulith dependency setup causes BOM not being added

odrotbohm:
The removal of the core dependency causes the BOM not being added as that's contained in the dependency declaration. Manually adding the BOM (build.boms().add(…)) adds the BOM segment but does not declare the version property that's needed to resolve it properly.


mhalbritter:
Thanks Ollie!

PR #1491, opened 2024-05-30
Upgrade to Spring Cloud GCP 5.4.1

JoeWang1127:

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -113,7 +113,7 @@ initializr:
         additionalBoms: [ spring-cloud ]
         mappings:
           - compatibilityRange: "[3.2.0,3.3.0-M1)"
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @JoeWang1127. I see that 5.4.0 upgrades to Micrometer Tracing 1.3.0 but continues to use Spring Boot 3.2.x. This may be a problem as it's Spring Boot 3.3.x that's tested and known to be compatible with Micrometer Tracing 1.3.

JoeWang1127:
Hi @wilkinsona, thanks for the reply.

We decide to release a patch (5.4.1) to remove the micrometer tracing from pom so that this dependency is managed by spring boot.

I'll update this PR when the release is finished.

mhalbritter:
Thanks!

PR #1509, opened 2024-06-12
Update description for vaadin entry

ZheSun88:
Since Vaadin 24.4, we provide original vaadin and hilla in this unified platform.


mhalbritter:
Thanks!

Issue #1510, opened 2024-06-12
Cursor pointer is not displayed for theme switcher in mobile view


sanajitjana:
When switching between dark and light themes on the mobile header of the start.spring.io website, the theme change icon does not display a cursor pointer. This can be confusing for users as it does not indicate that the icon is clickable.

Proposed Solution:
Add CSS to ensure the cursor changes to a pointer when hovering over the theme change icon on mobile header.

I have already implemented the code for this enhancement and would like to open a Pull Request. Can I proceed with submitting a PR for this issue?

Thank you for considering my contribution.

mhalbritter:
Hello,

yes, please submit a PR. Thanks!

mhalbritter:
Superseded by #1512.

PR #1512, opened 2024-06-13
Add cursor pointer for theme switcher in mobile view

sanajitjana:
#1510

### DIFF
--- a/start-client/src/styles/_main.scss
+++ b/start-client/src/styles/_main.scss
@@ -196,6 +196,9 @@ body {
       border: 0;
       padding: 4px;
       outline: none;
+      &:hover {
+        cursor: pointer;
+      }
     }
     svg {
       display: block;


start-client/src/styles/_main.scss
```diff
@@ -196,6 +196,9 @@ body {
       border: 0;
       padding: 4px;
       outline: none;
+      &:hover {
```
mhalbritter, line 199: The button in the non-mobile view has `cursor: pointer;` without the hover pseudoclass on it. Can we remove the pseudoclass here or did I overlook something?

mhalbritter:
Hey, thanks for the PR. I've added a comment.

start-client/src/styles/_main.scss
sanajitjana: Yes, we can directly use this cursor: pointer; under outline: none;
It will do the same job.

mhalbritter: Great, then please let's do that for consistency.

mhalbritter:
Thanks!

PR #1524, opened 2024-06-25
Fix history icon hover and transition on dark theme

sanajitjana:
Fix history icon hover and transition on dark theme


oodamien (approved):

oodamien:
Thanks @sanajitjana

PR #1528, opened 2024-07-02
Automatically add Spring Modulith event externalization dependency

odrotbohm:
If a message broker *and* Spring Modulith are declared as project dependencies, the event externalization of Spring Modulith for the configured broker is now declared automatically as well. Also, the Spring Modulith Events API dependency is declared in compile scope to allow event types to be marked for externalization.


mhalbritter:
Thank you!

PR #1533, opened 2024-07-09
Remove `addGeneratedAnnotation` from DGS Codegen Maven configuration

erichaagdev:
This property adds a timestamp to the generated sources and provides little value.

The [default value of this property is `false` in the plugin](https://github.com/deweyjose/graphqlcodegen/blob/d02aed8d85e8f223e461a76372cbe556e76726fa/src/main/java/io/github/deweyjose/graphqlcodegen/Codegen.java#L149). 

This change also brings it more in line with Gradle which doesn't configure this property and [whose default value is also `false`](https://github.com/Netflix/dgs-codegen/blob/3bf84bc13af2e939979e53b76d760be6575a7638/graphql-dgs-codegen-gradle/src/main/kotlin/com/netflix/graphql/dgs/codegen/gradle/GenerateJavaTask.kt#L142).

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/dgs/DgsCodegenMavenBuildCustomizer.java
@@ -48,8 +48,7 @@ public class DgsCodegenMavenBuildCustomizer implements BuildCustomizer<MavenBuil
 										.add("schemaPaths",
 												(schemaPaths) -> schemaPaths.add("param",
 														"src/main/resources/graphql-client"))
-										.add("packageName", this.packageName + ".codegen")
-										.add("addGeneratedAnnotation", "true"))));
+										.add("packageName", this.packageName + ".codegen"))));
 		build.plugins()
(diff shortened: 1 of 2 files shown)


erichaagdev:
It seems I've triggered a discussion between @deweyjose and [Netflix/dgs-codegen](https://github.com/Netflix/dgs-codegen/issues/702) about adding a property to disabled setting the `@Generated` annotation's `date` field. 

If that were to happen and the `date` field is not populated by default, then the `addGeneratedAnnotation` configuration option could remain enabled.

wilkinsona:
Thanks, @erichaagdev. I've subscribed to the dgs-codegen issue. Let's see how that pans out before doing anything here.

erichaagdev:
`dgs-codegen` has [merged a change](https://github.com/Netflix/dgs-codegen/issues/702#issuecomment-2240643887) to add a new `disableDatesInGeneratedAnnotation` option.

Once a new version of `dgs-codegen` is released, I'll open a corresponding PR to the Maven plugin. Once a new version of that is released, I'll create a PR here to enable `disableDatesInGeneratedAnnotation` by default.

As this could take some time, I'll close this PR for now.

PR #1541, opened 2024-07-27
HELP.md contains broken links to reference docs

ngocnhan-tran1996:
I update reference links in `HELP.md`. Please tell me if I miss something

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/build/gradle/GradleBuildSystemHelpDocumentCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/build/gradle/GradleBuildSystemHelpDocumentCustomizer.java
@@ -20,6 +20,8 @@ import io.spring.initializr.generator.project.ProjectDescription;
 import io.spring.initializr.generator.spring.documentation.HelpDocument;
 import io.spring.initializr.generator.spring.documentation.HelpDocumentCustomizer;
 import io.spring.initializr.generator.version.Version;
+import io.spring.initializr.generator.version.VersionParser;
+import io.spring.initializr.generator.version.VersionRange;
 
 /**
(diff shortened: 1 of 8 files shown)


mhalbritter:
Thank you very much and congratulations on your first contribution :tada:!

ngocnhan-tran1996:
@mhalbritter 

Thanks for your review.

I also find some old link in `application.yml`. Is there any way to add new item with same id or it will be changed when Spring 3.2.x goes to End of Support?

https://github.com/spring-io/start.spring.io/blob/722e50dd9e649a71f15e231dfb4cbfaf04b1a4fb/start-site/src/main/resources/application.yml#L221-L240

mhalbritter:
I have to look into this. FYI: When https://github.com/spring-projects/spring-boot/issues/41632 is resolved, this won't be a problem.

mhalbritter:
It's possible by using a custom `HelpDocumentCustomizer`, but there are a lot of such links in the `application.yaml`. I don't think it's worth the time to fix all of them for 3.2.x and 3.3.x. I hope https://github.com/spring-projects/spring-boot/issues/41632 resolves soon, which solves that problem at a central place.

And when 3.4.x is released in November, 3.2.x will drop out of OSS support, and then we can remove the remaining `if`s in the code.

ngocnhan-tran1996:
Thanks for your information. But I think [Link.java](https://github.com/spring-io/initializr/blob/main/initializr-metadata/src/main/java/io/spring/initializr/metadata/Link.java) need to be improved for multiple version document, add field `comparisonRange` for this purpose or something like that?

mhalbritter:
Yeah, if we would like to specify the version range for links in the metadata YAML. But as I said, I'd not go down that route and instead wait for Spring Boot to fix the redirects.

PR #1554, opened 2024-08-15
feat/new-test-DockerServiceResolver

ferclager:
Add a new test for DockerServiceResolver interface to cover all lines of code.


mhalbritter:
Hello,

thanks for the PR. However I don't think this test adds enough value to be included. If that needs do be tested, it would be better to test `SimpleDockerServiceResolver` instead.

ferclager:
Thanks for your comment. I will have in mind the class that you suggest to me.

PR #1582, opened 2024-09-11
Generate test app when spring cloud azure storage is selected

eddumelendez:
Spring Cloud Azure 5.16.0 supports Testcontainers and Docker Compose
Service Connection for Blob Storage and Queue.

* Add `spring-cloud-azure-docker-compose` or `spring-cloud-azure-testcontainers` dependency
* Configure `docker-compose.yaml` with image `mcr.microsoft.com/azure-storage/azurite:latest`
* Configure `GenericContainer` with image `mcr.microsoft.com/azure-storage/azurite:latest`


mhalbritter:
Hello! Thanks for the PR. Currently, we have multiple Azure dependencies:

image

Testcontainers and Docker Compose support is added based on the presence of a dependency with the group id `com.azure.spring`. AFAIK azurite only provides storage APIs. Should we instead make that conditional on the "Azure Storage" dependency instead of all Azure products?

eddumelendez:
Hi @mhalbritter! I have updated the code to only add the dependencies when `azure-storage` is selected along with `docker compose` and `testcontainers`.

mhalbritter:
Thanks a lot!

PR #1619, opened 2024-10-15
Add Spring AI document readers to initializr

### LINKED ISSUE #1618: Spring AI Document Readers
Add the Spring AI Document readers to the umbrella of Spring AI dependencies provided by the Initilizr. 

The [Document Readers](https://docs.spring.io/spring-ai/reference/api/etl-pipeline.html#_documentreaders) provide a common mechanism to load files of different formats into Document objects stored in the Vector Database. 
So far we provide support for Tika, PDF and Markdown documents types. 

The Document Reader are part of the same BOM like the reset of the Spring AI dependencies, but they are not starters.

tzolov:
- Include Markdown, Tika, and PDF document readers
 - Add descriptions and reference links for each reader
 - Set Spring AI BOM and disable starters

 Resolves #1618

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -1678,6 +1678,36 @@ initializr:
           links:
             - rel: reference
               href: https://docs.spring.io/spring-ai/reference/api/vectordbs/weaviate.html
+        - name: Markdown Document Reader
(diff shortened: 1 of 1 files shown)


mhalbritter:
Thanks @tzolov !

PR #1623, opened 2024-10-17
Update node to the latest LTS version

ferclager:
Update to the latest node v16 released on 2023-08-08.


mhalbritter:
Hey, thanks for the PR. I've noticed that we use an unsupported version of node and have updated node to 20.18.0, which is the latest LTS.

PR #1640, opened 2024-11-08
Replace Collectors.toList() with toList()

eddumelendez:


mhalbritter:
Thanks @eddumelendez !

PR #1650, opened 2024-11-25
Upgrade Spring Boot Admin to 3.3.6 and 3.4.0

SteKoe:
Hey, let’s go! :) Spring Boot Admin 3.3.6 version has arrived as well as a compatible release for Spring Boot 3.4.0!


wilkinsona:
Thanks, @SteKoe, and congratulations on the releases.

PR #1664, opened 2024-12-05
Add ability to bookmark projects

### LINKED ISSUE #1487: Add "Favorite" action to history
As the history is limited to the last N items, it would be nice to have a way to favorite history items so that they are always kept.

oodamien:
Resolves #1487


mhalbritter:
Thanks very much @oodamien !

PR #1676, opened 2024-12-18
Add help section and Maven plugin when azure-support is selected

Netyyyy:
When select azure-support:
 - Add `Deploy to Azure` section in Help.md
 - Add `azure-container-apps-maven-plugin` in pom


mhalbritter:
Thanks @Netyyyy !

Issue #1675, opened 2024-12-17
Use @generated=omit when configuring gRPC plugins


alexanderankin:
when you add grpc in the initializr it adds this to build.gradle

```gradle
protobuf {
  protoc {
    artifact = 'com.google.protobuf:protoc'
  }
  plugins {
    grpc {
      artifact = 'io.grpc:protoc-gen-grpc-java'
    }
  }
  generateProtoTasks {
    all()*.plugins {
      grpc {
        option 'jakarta_omit'
      }
    }
  }
}
```

but it should also include this other option to be forwards compatible, based on these release notes - https://github.com/grpc/grpc-java/releases/tag/v1.64.0

```gradle
// ...
        option '@generated=omit'
// ...
```

the simpler jakarta_omit option was added at this time - https://github.com/grpc/grpc-java/releases/tag/v1.63.0

i have created a patch for this fix here - https://github.com/spring-io/start.spring.io/compare/main...alexanderankin:start.spring.io:fix/future-compatible-grpc-option?expand=1

I initially opened this issue in the wrong repo (here - https://github.com/spring-io/initializr/issues/1601)

mhalbritter:
Thanks for letting me know. Do you mind open a PR with your changes? And please remove the comments. We can drop `jakarta_omit` once spring-grpc has upgraded `protoc-gen-grpc-java` to `1.64`.

alexanderankin:
Sure thing, opened https://github.com/spring-io/start.spring.io/pull/1682

mhalbritter:
Superseded by https://github.com/spring-io/start.spring.io/pull/1682.

PR #1682, opened 2024-12-18
Fix/Use future compatible grpc/protoc option

alexanderankin:
introduces new flag per #1675 but does not remove old flag yet


mhalbritter:
Thanks @alexanderankin !

PR #1702, opened 2025-01-17
Update application.yml for new snapshots of Spring gRPC

dsyer:


mhalbritter:
Instead of updating to 0.4.0-SNAPSHOT, we decided to upgrade to 0.3.0 instead.

PR #1710, opened 2025-01-26
Allow removal of dependencies from the dependencies selection UI

aloussase:
Users might misclic or change their mind about adding a dependency. Having to close the dialog in order to remove the dependency is unnecessary extra work.

This makes it so that the selected dependencies list is shown inside the dialog in a non-obstrusive way so that users can see what they've chosen so far and potentially remove them.

https://github.com/user-attachments/assets/a3b5955b-c350-432b-8b7b-4147cd0986d2

### DIFF
--- a/start-client/src/components/common/dependency/Dialog.js
+++ b/start-client/src/components/common/dependency/Dialog.js
@@ -256,6 +256,9 @@ function Dialog({ onClose }) {
                       }}
                       onRemoved={(item) => {
                         setSelectedDeps(selectedDeps.filter(id => id !== item.id))
+                        if (item.valid && !multiple) {
+                          onClose()
+                        }
                       }}
                       onSelect={() => {
                         setSelected(index)
@@ -286,6 +289,9 @@ function Dialog({ onClose }) {
                             }}
                             onRemoved={(item) => {
                               setSelectedDeps(selectedDeps.filter(id => id !== item.id))
+                              if (item.valid && !multiple) {
+                                onClose()
+                              }
                             }}
                             onSelect={i => {
                               setSelected(i)
--- a/start-client/src/components/common/dependency/Dialog.js
+++ b/start-client/src/components/common/dependency/Dialog.js
@@ -48,6 +48,13 @@ function Dialog({ onClose }) {
     })
   }
 
+  const remove = id => {
+    dispatch({
+      type: 'REMOVE_DEPENDENCY',
+      payload: { id: id },
+    })
+  }
+
   useEffect(() => {
     const jsSearchUp = new JsSearch.Search('name')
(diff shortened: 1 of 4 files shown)


mhalbritter:
Hello,

first of all, thanks for this PR! We talked about this in our meeting and we agree that the current behavior of the list is not optimal.

We looked at your design in your video, and while it improves things, we'd like to propose something:

what about, instead of the green bubbles showing up on the top when you select a dependency, the dependency would stay in the list but get a different colored background. That way the user would see all selected dependencies (because they are colored), is able to de-select them by clicking on them again and there's no problem with scrolling when a lot of dependencies are selected.

That's essentially the same behavior as multi-select lists in operating systems and a lot of UIs: Ctrl+Click a deselected entry to select it and the background color indicates that it's selected. Ctrl+Click it again while selected deselects it, restoring the original color.

We think that this would work here, too.

Would you be willing to implement that idea?

aloussase:
https://github.com/user-attachments/assets/4686cbd9-fd85-4532-9923-b4aa86599467

Hello, @mhalbritter 

I have made the requested changes. Let me know what you think :)

mhalbritter:
This looks awesome! I found a small issue:

1. Open dependency picker UI, select dependency, UI closes
2. Open dependency picker UI again, dependency is now selected, click the dependency
3. The dependency is removed, however the UI doesn't close. I would expect the UI to close, to mirror the behavior when adding the dependency

mhalbritter:
And another thing: the search field in the dependencies UI only displays non-added dependencies. If a dependency has been added, i can't find it again. I'd expect that the search also shows already added dependencies, which are then shown with a different colored background.

aloussase:
@mhalbritter I have resolved the issues you mentioned

Regarding the last point, I could not reproduce, but I made it so that when pressing enter
on an added dependency, it is removed, which was missing before.

mhalbritter:
Very nice, thanks for working on this! I can't reproduce the strange behavior with the keyboard navigation either. Maybe something else you changed fixed that.

Now, to get this merged, you only need to apply DCO to your commits, see [the contribution guide](https://github.com/spring-io/start.spring.io/blob/main/CONTRIBUTING.adoc#include-a-signed-off-by-trailer). [This blog post](https://spring.io/blog/2025/01/06/hello-dco-goodbye-cla-simplifying-contributions-to-spring) has more information.

aloussase:
Hello @mhalbritter , 

I have added the DCOs. Didn't know about them, very interesting!

Thanks,
Alexander

mhalbritter:
This is now merged and will be deployed soon. Thank you very much @aloussase for your work, this is a really nice feature!

PR #1733, opened 2025-02-25
Update Spring Modulith versions

odrotbohm:
Also added version mappings for the Boot 3.5 generation. Arbitrarily chose 3.6.0-M1 as forward delimiter. Please adapt if that was the wrong choice.


mhalbritter:
`1.4.0-M2` isn't on central, right? It's only in spring milestones?

PR #1742, opened 2025-03-03
Disable dates in @Generated annotations for DGS Codegen Maven

erichaagdev:
This is a reimplementation of https://github.com/spring-io/start.spring.io/pull/1533 now that the plugin offers the `disableDatesInGeneratedAnnotation` configuration option. These dates prevent byte-for-byte reproducibility of the generated sources.


mhalbritter:
Thanks @erichaagdev !

PR #1762, opened 2025-03-21
Update Timefold compatibility range

zepfred:


mhalbritter:
So just to be clear, you have CI in place which tests against 3.3.x and 3.4.x, correct?

zepfred:
> So just to be clear, you have CI in place which tests against 3.3.x and 3.4.x, correct?

Yes, we do.

triceo:
Hello @mhalbritter!
This PR is to keep consistency between two places in the file which reference Timefold Solver. The other place already has this exact version range.

To answer your question though - we have CI for latest Spring, therefore 3.4.x.
If absolutely necessary, we can add one for 3.3.x as well.

PR #1782, opened 2025-04-25
Upgrade Spring gRPC to 0.8.0

dsyer:


mhalbritter:
@dsyer Spring gRPC is still listed as experimental. Should I remove that?

image

dsyer:
> @dsyer Spring gRPC is still listed as experimental. Should I remove that?

Yes please. Good catch.

PR #1789, opened 2025-04-30
Upgrade to Spring AI 1.0.0-M8

ilayaperumalg:


start-site/src/main/resources/application.yml
```diff
@@ -83,8 +83,8 @@ initializr:
         artifactId: spring-ai-bom
         versionProperty: spring-ai.version
         mappings:
-          - compatibilityRange: "[3.4.0,3.5.0-M1)"
-            version: 1.0.0-M7
+          - compatibilityRange: "[3.4.0,3.5.0-M3)"
```
snicoll, line 86: I don't understand this. Spring Boot is at 3.5.0-RC1 at the moment so you won't gain anything by changing the range mid-flight like this.

What are you trying to convey? Is Spring AI M8 compatible with Spring Boot 3.5 or not?

ilayaperumalg: I see. Spring AI M8 was tested against Spring Boot 3.5.0-M3 and hence we updated the range. Perhaps this should be 3.5.0-RC1 to include the support for 3.5.0-M3?

ilayaperumalg: ok, thanks for clarifying. yes, It is supposed to support Spring Boot 3.4 as well.

ilayaperumalg: Discussed with Stephane and I am currently running the tests against Spring Boot 3.5.0-RC1. With that, we can update the range accordingly.

PR #1805, opened 2025-05-22
Switch gRPC plugin to io.github.ascopes:protobuf-maven-plugin

dsyer:
The samples at Spring gRPC have all been switched to this other plugin because the old one has fallen out of maintenance. It works on the command line right now, but Eclipse/VSCode users will want the <?m2e ...?> hints to drive the build. See https://github.com/spring-io/initializr/pull/1650


dsyer:
I forgot about the proto file path. I think we should use the default, so (since we don’t ship samples from initializr with proto files) there’s nothing to do? Or can we create the empty directory as a hint?

I didn’t know the os classifier wasn’t needed. Maybe we should verify on all platforms?

dsyer:
N.B. I don't want to merge this yet until we either add an `<?m2e ...?>` hint (requires https://github.com/spring-io/initializr/pull/1650) or the plugin gets a new release (3.2.4 will hopefully have the metadata OOTB).

mhalbritter:
Yeah, we can add processing instructions until the plugin authors release their new version.

dsyer:
That should work for a standalone sample but it breaks if the app is not the top level in the Maven reactor. We can see if it can be fixed in the plugin (I opened an issue). I prefer to just use the default here for now,

mhalbritter:
Thanks @dsyer !

PR #1806, opened 2025-05-22
remove extraspace - ServiceConnections.java

Aamirmalick:
remove extraspace in the code


wilkinsona:
Thanks for the PR, but these blank lines are intentional. They provide symmetry with the whitespace around the matching opening braces.

PR #1808, opened 2025-05-23
Update Vaadin's compatibility range to include Spring Boot 3.5

manolo:

PR #1810, opened 2025-05-26
chore: update Vaadin compatibilityRange for supported versions

ZheSun88:
- Vaadin 24.5 has been EOL for a while, starting from the next Vaadin minor, the minimal supported spring boot version is 3.4. so we decided to drop the compatibilityRange [3.3.0, 3.4.0-M1)
- update the compatibilityRange for Vaadin 24.7 to include springboot 3.5 support


ZheSun88:
we can close #1808

PR #1829, opened 2025-06-04
Update Sentry compatibility range to support Spring Boot 3.5.x

lcian:
Sentry supports Spring Boot 3.5.x. 
This PR updates the compatibility range so that this library can be added to projects using 3.5.x.


mhalbritter:
Thanks @lcian !

PR #1832, opened 2025-06-07
Upgrade netflix-dgs to 10.2.1 for Spring Boot 3.5

joelpelaez:
Add new version of netflix-dgs 10.2.1 that includes compatibility with Spring Boot 3.5 Release series.

This change keeps older versions for previous Spring Boot Releases.


mhalbritter:
Thanks @joelpelaez !

PR #1855, opened 2025-07-06
Drop old 'jakarta_omit' option

breun:
The `jakarta_option` was only supported by a single grpc-java release and quickly replaced by `@generated=omit` (grpc/grpc-java#11086), which is used by the current Spring gRPC version.

Also see https://github.com/spring-projects/spring-grpc/pull/225, where this option was removed from the Spring gRPC sample projects.


onobc:
I have seen inconsistent behavior between Maven and Gradle wrt to these options. Let's just be sure that both Maven and Gradle work as expected when we remove this option.

breun:
Might be best to let CI verify the sample projects then?

breun:
The `jakarta_omit` option only existed in grpc-java 1.64.0 and was quickly replaced with `@generated=omit` in the next release. It doesn't have any effect unless you use that one particular older version of grpc-java, but that version is not used by projects generated via start.spring.io.

mhalbritter:
I've tested it with Maven and Gradle.

mhalbritter:
Thanks @breun !

PR #1964, opened 2025-10-30
Scroll to last added dependency on dialog reopen

n4j1Br4ch1D:
Ensure the dependency selection dialog automatically scrolls so the
last added dependency is aligned at the top of the list when reopened.
This improves usability when adding multiple dependencies in sequence.

scrollToLastAddedDependency

### DIFF
--- a/start-client/src/components/common/dependency/Dialog.js
+++ b/start-client/src/components/common/dependency/Dialog.js
@@ -119,6 +119,20 @@ function Dialog({ onClose }) {
     setSelected(0)
     setQuery('')
     textFocus()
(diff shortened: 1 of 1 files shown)


wilkinsona:
I think I'd find this frustrating if I wasn't adding dependencies in sequence. If I was, I'd prefer to hold down CTRL or CMD to keep the dialog open.

mhalbritter:
Thanks for the PR, but we don't like this behavior. We already have a way for adding dependencies in sequence and I don't see the value in scrolling to the last selected one.

PR #1969, opened 2025-11-06
Enhance JavaVersionProjectDescriptionCustomizer to handle Gradle–JDK incompatibility (#1937)

mukeshlilawat1:
### Summary
This PR improves project generation robustness by introducing an additional
Gradle–JDK compatibility validation in `JavaVersionProjectDescriptionCustomizer`.

### Details
- Detects when an older Gradle version (e.g., 7.x used by Boot < 3.0) is combined
  with newer Java versions (22+).
- Logs a warning and automatically falls back to a supported JDK version.
- No behavioral impact for Maven builds or already compatible setups.

### Impact
Prevents project generation failures for Gradle users selecting JDK versions
that Gradle does not yet support.

### Testing
Verified with Gradle and Maven builds for multiple Boot and JDK versions.
All existing tests pass successfully.

PR #1968, opened 2025-11-05
Enhance JavaVersionProjectDescriptionCustomizer robustness and fallback handling

mukeshlilawat1:
### Summary
This PR improves the robustness of the `JavaVersionProjectDescriptionCustomizer` by
introducing safer handling for invalid or blank JVM versions and preventing potential
null pointer scenarios.

### Changes
- Added null-safety checks for `MutableProjectDescription` and `Language`.
- Introduced fallback to minimum supported Java version when blank or invalid input is provided.
- Optimized `updateTo()` to avoid unnecessary re-creation of language instances.
- Improved code readability and logical flow without altering existing behavior.

### Impact
This enhancement ensures consistent project generation behavior for all supported
Spring Boot and Java version combinations. It introduces no breaking changes.

### Testing
Verified locally with multiple JVM versions (8–25) and existing unit tests.  
No regressions observed; build and tests passed successfully.

PR #1972, opened 2025-11-06
Provide user indication when Java version is downgraded or adjusted (#250)

### LINKED ISSUE #250: Provide some indication to the user that the Java version has been or will be downgraded
There are a few scenarios when the Java version is silently downgrade. For example, when Kotlin is selected as the language, Java 8 will always be used. Another example in the future will be when Java 13 is offered. Gradle won't support Java 13 until 6.0 so a downgrade may be necessary in the interim.

The silent downgrade has caused some confusion in the past and continues to do so. See https://github.com/spring-io/initializr/issues/739, for example. We should consider making the downgrade clearer to the user. One solution was originally being tracked in https://github.com/spring-io/initializr/issues/773 until it was repurposed.

There are two mains approaches that have been thought of thus far:

1. Update the UI to prevent combinations that would result in a downgrade
2. Add a note in the `HELP.md` file that explains what has happened and why

We may even need a hybrid of the two as the first solution won't help for projects generated using a client other than the web UI.

mukeshlilawat1:
### Overview
This PR enhances the `JavaVersionProjectDescriptionCustomizer` to make automatic
Java version adjustments visible to users. Previously, when Spring Initializr
downgraded or upgraded the Java version for compatibility, it happened silently,
providing no user indication.

This change fulfills the request from issue #250 by logging a clear and contextual
message whenever the requested Java version is automatically changed due to:
- Unsupported legacy Java versions (e.g., 1.7, 1.8)
- Versions too new for the chosen Spring Boot release
- Invalid or unspecified Java versions
- Gradle–JDK incompatibility scenarios (retains fix from #1969)

---

### Key Enhancements
- Added descriptive messages when Java version is upgraded/downgraded.
- Introduced `logAdjustment()` to attach or log compatibility messages.
- Extended `updateTo()` to include user-facing indication for version changes.
- Preserved existing validation logic and Gradle–JDK compatibility handling.
- Backward compatible and safe for all existing configurations.

---
### Example Behavior
[INFO] Java version 26 not yet supported. Downgraded to 25.
[INFO] Gradle build incompatible with Java 22. Adjusted to 21.
[INFO] Java version 7 too old. Upgraded to supported version 17.

---

### Testing
- Verified with multiple Boot versions (2.7.x, 3.x) and Gradle build systems.
- Added unit test to confirm downgrade detection.

---

### Related Issues
- Closes #250  
- Also retains logic for #1969 (Gradle–JDK compatibility)

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/project/JavaVersionProjectDescriptionCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/project/JavaVersionProjectDescriptionCustomizer.java
@@ -27,7 +27,9 @@ import io.spring.initializr.generator.version.Version;
 
 /**
  * Validate that the requested java version is compatible with the chosen Spring Boot
- * generation and adapt the request if necessary.
+ * generation and adapt the request if necessary. Also includes a compatibility
+ * check between selected Gradle build versions and Java versions to avoid
+ * unsupported build setups.
  *
  * @author Stephane Nicoll
  * @author Madhura Bhave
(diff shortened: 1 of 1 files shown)


mhalbritter:
Hello - this PR is very similar to https://github.com/spring-io/start.spring.io/pull/1968. I'm not sure if this is an accident, but if not, please do not open PRs for code changes which already have been declined.

PR #1978, opened 2025-11-10
Fix Spring Guides branch name

ngocnhan-tran1996:


mhalbritter:
Thanks @ngocnhan-tran1996 !

PR #1985, opened 2025-11-13
Upgrade to Spring AI 1.1.0

ilayaperumalg:
Updated compatibility range for AI models and added support for new AI models and vector databases in the application configuration.

### DIFF
--- a/start-site/src/main/resources/application.yml
+++ b/start-site/src/main/resources/application.yml
@@ -89,8 +89,8 @@ initializr:
         artifactId: spring-ai-bom
         versionProperty: spring-ai.version
         mappings:
-          - compatibilityRange: "[3.4.0,4.0.0-M1)"
-            version: 1.0.3
+          - compatibilityRange: "[3.5.0,4.0.0-M1)"
+            version: 1.1.0
       spring-cloud:
         groupId: org.springframework.cloud
         artifactId: spring-cloud-dependencies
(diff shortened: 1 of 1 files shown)


mhalbritter:
Hey, thanks for the PR. Do you want to get users, if they select Boot 3.4.x, the 1.0.3 version of AI? Because right now, this PR would make Spring AI unusable with Spring Boot 3.4.x, as far as I can tell.

mhalbritter:
But maybe we can live with that, as Boot 3.4.x will disappear anyway when it's out of OSS support, which will happen in December.

ilayaperumalg:
@mhalbritter thanks for the quick follow-up! 

yeah, we can only show 1.1.0 as 1.0.x will not be actively maintained going forward as well.

ilayaperumalg:
Do you think we can remove the `buildWithSpringBootAndTestcontainersSpringAiModule` test from TestcontainersProjectGenerationConfigurationTests as it only checks for Spring Boot 3.4 compatibility and I see the test SpringAiTestcontainersProjectGenerationConfigurationTests covers Spring Boot 3.5 compatibility?

snicoll:
>  as 1.0.x will not be actively maintained going forward as well.

Spring AI 1.0.x OSS support ends next June. Shouldn't we keep it around as long as Spring Boot 3.4.x is around?

ilayaperumalg:
> Spring AI 1.0.x OSS support ends next June. Shouldn't we keep it around as long as Spring Boot 3.4.x is around?

Ideally that would be the case. Given Spring AI 1.1.x also falls along the same line with Spring Boot 3.5.x, we are planning to actively maintain Spring AI 1.1.x while keeping 1.0.x for any critical fixes alone. We just switched Spring AI 2.0.0 as the next major. Moreover, there have been significant feature set changes between Spring AI 1.0.x and 1.1.x including many deprecations and retired items we don't have any control over. As a user, someone would lean towards switching to 1.1.x at this point.

mhalbritter:
Thanks for the PR @ilayaperumalg, that was very helpful.

We now have support for Spring AI 1.0.3 (for Boot 3.4.x) and Spring AI 1.1.0 (for Boot 3.5.x).

ilayaperumalg:
@mhalbritter thank you for taking care of including both Spring AI 1.0.3 and 1.1.0!

Issue #1984, opened 2025-11-13
Website has to be scrolled vertically on smaller screens


mhalbritter:
Since we added the configuration format radio buttons, the website needs to be scrolled vertically on smaller screens (such as a 13" MacBook).

We could maybe fix that by moving artifact id and group id into one row, or drop the description.

sri1873:
<img width="908" height="451" alt="Image" src="https://github.com/user-attachments/assets/89e4d0e2-7196-47aa-920b-2a518fa82a99" />

sri1873:
Hey @mhalbritter, is this what you had in mind?

I have temporarily disabled description in the UI and wrapped **groupId** and **artifactId** into a single div. I didn’t have to do any CSS changes for now, but if that is ok, I’ll work on this and run it across multiple screen sizes to confirm nothing breaks.

mhalbritter:
Hello,

thanks for looking at this! It's what I had in mind, except that we can leave the description in there, if we save enough space (and i think we do) by moving group and artifact id next to each other.

mhalbritter:
Closing in favor of https://github.com/spring-io/start.spring.io/pull/2010.

PR #2010, opened 2025-12-04
Website has to be scrolled vertically on smaller screens

sri1873:
**Changes Made  for #1984**

1.  Put GroupId and artifactId into a single div. 
2. Reduced padding from 0.6rem to 0.45rem to fit them on one screen.

image

### DIFF
--- a/start-client/src/components/common/builder/Fields.js
+++ b/start-client/src/components/common/builder/Fields.js
@@ -109,7 +109,7 @@ function Fields({
               )}
             </Control>
             <Control text='Project Metadata'>
-              <div className="control control-inline">
+              <div className="control-row">
                 <FieldInput
                   id='input-group'
                   value={get(values, 'meta.group')}
--- a/start-client/src/styles/_main.scss
+++ b/start-client/src/styles/_main.scss
@@ -278,7 +278,7 @@ hr.divider {
 // Control
 
 .control {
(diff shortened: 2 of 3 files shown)


mhalbritter:
Thanks for the PR - this looks okay-ish on smaller screens, but on my screen in full screen it looks like this:

image

There's a lot of "wasted" space after the "Artifact" textbox, and before the "Artifact" label. Adjusting the padding of all "control" classes also alters the whole layout, not only for the two text fields which are now next to each other.

I also think the spacing between "Group" and "Name" is bigger than the spacing between "Name" and "Description".

sri1873:
image

New Changes

1. Reverted "control" classes padding to the original.
2. Adjusted the width of  text input fields to maintain alignment, as well as use up space

Between 1000px and 1200px, the input fields of "Group" and "Artifact" become too small, to avoid that, in those screen sizes the fields will go back to occupying the entire row.

Please let me know if something else needs to be changed

mhalbritter:
Thanks @sri1873 !

mhalbritter:
I [made a smaller change](https://github.com/spring-io/start.spring.io/commit/ef0c52e83f6523f95c3072ddc5591296b91925cd) to get rid of the `nth-child(2)` selector by adding an explicit class.

PR #2011, opened 2025-12-04
gRPC Maven Plugin uses deprecated protoc configuration

breun:
The release notes of [protoc-maven-plugin 4.0.0](https://github.com/ascopes/protobuf-maven-plugin/releases/tag/v4.0.0) say:

> `<protocVersion>` is now an alias for `<protoc>` rather than vice versa. Users can continue to use the old name until v5 but are encouraged to update.

This pull request replaces the deprecated `protocVersion` element with the more future-proof `protoc` which should stay working when upgrading to version 5.x in the future.


mhalbritter:
Thanks @breun !

Issue #1795, opened 2025-05-09
Deal with the warnings from sass


mhalbritter:
After a migration from `node-sass` to `sass` warnings appeared, and we should deal with them. You can see them when running `mvn package` in the `client-site` folder.

Ahmed1571998247:
150

sri1873:
Hi @mhalbritter  can i work on this issue ?

mhalbritter:
Sure, thank you!

mhalbritter:
Superseded by https://github.com/spring-io/start.spring.io/pull/2020.

PR #2020, opened 2025-12-15
Deal with the warnings from sass

sri1873:
This PR is resolves the issue #1795 

- Migrated [deprecated @import](https://sass-lang.com/documentation/breaking-changes/import/) usage to modern `@use`

- Since `@use` does not provide global scope,  shared variables are explicitly imported using `@use 'variables' as *` to avoid widespread refactors across existing styles

- Replaced [deprecated `lighten()` / `darken()`](https://sass-lang.com/documentation/breaking-changes/color-functions/) calls with `color.adjust()`
 ```
 lighten($c, 10) → color.adjust($c, $lightness: 10%)
 darken($c, 10) → color.adjust($c, $lightness: -10%)
 ```
- **Note:** Not all Sass warnings are resolved: one remaining deprecation warning originates from the third-party `hamburgers` library imported from node_modules, which is outside the scope of this change

### DIFF
--- a/start-client/src/styles/_hamburgers.scss
+++ b/start-client/src/styles/_hamburgers.scss
@@ -16,7 +16,7 @@ $hamburger-padding-y: 8px;
 $hamburger-layer-width: 26px;
 $hamburger-layer-height: 3px;
 $hamburger-layer-spacing: 6px;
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks @sri1873 !

PR #2021, opened 2025-12-16
Update tanzu-scg-extensions to 1.0.1

spikymonkey:
Updates `tanzu-scg-extensions` to `1.0.1`, which is being released today.


mhalbritter:
Thanks @spikymonkey !

PR #2025, opened 2025-12-18
Upgrade to Netflix DGS 11.0.0 and DGS Codegen 8.3.0

bclozel:
The new Netflix DGS generation supports Spring Boot 4.0. This also upgrades the DGS Codegen for Gradle 9+ compatibility.


mhalbritter:
Thanks @bclozel !

PR #2027, opened 2025-12-19
Update test after Spring Boot 4.0.1 release

bclozel:
`ExtensionIntegrationTests` currently fails because it is checking for the exact "4.0.0" version for the currently supported version - "4.0.1" recently got out and the test breaks as a result.


mhalbritter:
Thanks Brian. I've noticed the same and decided to drop the assertion in https://github.com/spring-io/start.spring.io/commit/57c62e58caad7703d0332ebdd63a91f9cb7b3c68.

PR #2030, opened 2025-12-23
Upgrade to Spring Modulith 2.0.1 and 1.4.6

odrotbohm:


odrotbohm:
Thanks for the quick turnaround! 🙇

Issue #1949, opened 2025-10-22
Add support for fixed ports in Docker Compose generation


mhalbritter:
For Grafana LGTM i wanted to generate '3000:3000' for the admin UI in the ports section of a Docker Compose file. We only support random ports, but in this case, having a fixed port for the UI (which the user can then open in the browser) would make sense.

Alternative would be that Spring Boot prints out a link to the Admin UI port on startup.

EduardoRangelG:
Hi @mhalbritter, I'd like to work on this! I've analyzed the current implementation and I see that to support fixed ports for Grafana, I first need to enhance the Spring Initializr library. 

**My plan is to:**

* Update ComposeService and ComposeFileWriter in Spring Initializr to support an optional hostPort while maintaining backward compatibility for existing random-port services.
* Update DockerService and SimpleDockerServiceResolver in start.spring.io to use this new capability specifically for Grafana.

Does this approach align with your vision for the fix? If so, could you please assign this to me?

mhalbritter:
Yes, that makes sense. Thank you.

EduardoRangelG:
Thanks. I am already working on the PR, and I hope to open it as soon as possible.

mhalbritter:
Superseded by https://github.com/spring-io/start.spring.io/pull/2040.

PR #2040, opened 2026-01-13
Use fixed port for Grafana LGTM

EduardoRangelG:
## Summary

Configure Grafana LGTM with fixed port 3000:3000 for web UI access while maintaining random ports for OTLP endpoints.

### Changes
- Add `portMapping()` methods to `DockerService.Builder` to support both random and fixed ports
- Update `grafanaLgtm()` service to use fixed port 3000:3000 for the web UI
- Maintain backward compatibility with existing `ports()` methods

## Why Fixed Port for Grafana?

The Grafana web UI needs a predictable port (3000) so users can easily open it in their browser after starting the Docker Compose services. Other ports (4317, 4318 for OTLP) remain random as they don't require direct browser access.

Closes spring-io/start.spring.io#1949


mhalbritter:
Thanks @EduardoRangelG !

PR #2076, opened 2026-02-18
Generate test app when Spring AI Chat Memory Cassandra is selected

eddumelendez:


mhalbritter:
Thanks @eddumelendez !

Issue #2123, opened 2026-03-19
RabbitMQ support as of Spring Boot 4.1.0-M3 should use spring-boot-rabbitmq


snicoll:
See https://github.com/spring-projects/spring-boot/issues/49620 for context.

"Spring for RabbitMQ" and "Spring for RabbitMQ Streams" should migrate from `spring-boot-starter-amqp` to `spring-boot-starter-rabbitmq` as of Spring Boot `4.1.0-M3`.

mhalbritter:
Superseded by https://github.com/spring-io/start.spring.io/pull/2126.

PR #2126, opened 2026-03-20
Adapt to RabbitMQ behavior moving to spring-boot-starter-rabbitmq

snicoll:
RabbitMQ support has moved from "amqp" to "rabbitmq" as of Spring Boot 4.1.0-M3. As a result, the coordinates have moved accordingly as of this version.

This is a breaking API change as the "amqp" dependency id no longer resolve.

See gh-2123


mhalbritter:
See https://github.com/spring-projects/spring-boot/issues/49620#issuecomment-4097017563. Blocked on https://github.com/spring-projects/spring-boot/issues/49677.

PR #2168, opened 2026-04-22
Add infrastructure for Testcontainers generation code to add annotations

### LINKED ISSUE #2162: Add infrastructure for Testcontainers generation code to add @Ssl annotations
We need that for https://github.com/spring-io/start.spring.io/pull/2150.

kaiquevieirasoares:
Description
This PR enhances the ServiceConnection API to support custom annotations with attributes (parameters). This is a foundational change that enables the upcoming support for the @Ssl annotation required by newer versions of Elasticsearch (v8+), as originally discussed in #2162.

Changes

* Infrastructure: Introduced the AnnotationRequest record to store annotation class names alongside a Consumer<Annotation.Builder> for attribute customization.

 * API Evolution: Added withAnnotation(ClassName) and withAnnotation(ClassName, Consumer<Annotation.Builder>) to ServiceConnection. This approach elegantly delegates attribute type inference natively to the Initializr AST, allowing the seamless use of Strings, Booleans, and Integers without the need for manual string escaping.

* Tests: Added comprehensive integration tests covering these multiple attribute types to ensure correct generation across Java, Kotlin, and Groovy.

Fixes
Closes #2162

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/container/ServiceConnections.java
+++ b/start-site/src/main/java/io/spring/start/site/container/ServiceConnections.java
@@ -49,6 +49,14 @@ public class ServiceConnections {
 		return this.connectionsById.values().stream().sorted(Comparator.comparing(ServiceConnection::id));
 	}
 
+	/**
+	 * A single additional annotation to emit on the generated container method, and an
+	 * optional callback to customize the {@link Annotation.Builder}.
+	 *
+	 * @param className the fully qualified annotation to add
+	 * @param customizer the attribute configuration, or {@code null} to apply no
+	 * attributes
+	 */
 	public record AnnotationRequest(ClassName className, Consumer<Annotation.Builder> customizer) {
 		public AnnotationRequest {
 			customizer = (customizer != null) ? customizer : (builder) -> {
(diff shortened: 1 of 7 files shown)


start-site/src/main/java/io/spring/start/site/container/ServiceConnections.java
```diff
@@ -42,24 +45,45 @@ public Stream<ServiceConnection> values() {
 		return this.connectionsById.values().stream().sorted(Comparator.comparing(ServiceConnection::id));
 	}
 
+	public record AnnotationRequest(String className, Map<String, String> attributes) {
```
mhalbritter: This works for String attributes. What if i have a different type of attribute?

mhalbritter:
Thanks for the PR. I did a quick review and added a comment. Also, please add a Signed-off-by header to fix the failing DCO check, [see here](https://github.com/spring-io/start.spring.io/blob/main/CONTRIBUTING.adoc#include-a-signed-off-by-trailer).

start-site/src/main/java/io/spring/start/site/container/ServiceConnections.java
kaiquevieirasoares: Great catch! I initially used String so callers could pass raw literals (which is why I manually escaped strings in the tests). However, I completely agree that using Map<String, Object> provides a much cleaner API.
​I can update the AnnotationRequest to accept Object and handle the basic stringification inside the Contributors (e.g., automatically wrapping Strings in quotes, and rendering Numbers/Booleans as is).

​Does this approach sound good to you, or does the Initializr AST have a preferred built-in way to handle typed annotation attributes?

mhalbritter: Initializr already has a way, please look at the `io.spring.initializr.generator.language.Annotation.Builder#set` method.

kaiquevieirasoares: You're absolutely right! I've refactored this to use Consumer<Annotation.Builder>. This way, the Initializr AST handles the types natively (Strings, Booleans, Integers), making the API much cleaner and more flexible.

kaiquevieirasoares:
Thanks for the feedback, Moritz!

I've just force-pushed the updates:

  * Added the Signed-off-by header to the commit.

  * Refactored the AnnotationRequest to use Consumer<Annotation.Builder> as suggested. This handles type inference automatically and removes the need for manual string escaping.

  * Updated the integration tests to cover different attribute types (String, Boolean, Integer) across Java, Kotlin, and Groovy.

Let me know if everything looks good now!

mhalbritter:
Thanks @kaiquevieirasoares !

PR #2179, opened 2026-04-30
Let the 'More' button toggle the dropdown state

aalahyan3:
Improved UX by making the 'More' button labeled with favorite-add toggles the dropdown state instead of making it for open only

**Description**
This PR improves the UX of the 'More' button (associated with the favorite/share functionality).

**The Problem**
Previously, the button was configured to only trigger the open state of the dropdown. Clicking the button a second time while the dropdown was already visible would not close it (or would immediately re-open it), forcing users to click elsewhere on the screen to dismiss the menu.

**The Fix**
Updated the click handler in src/components/common/builder/Fields.js to properly toggle the state. The button now acts as a true toggle:

Click 1: Opens the dropdown.

Click 2: Closes the dropdown.

_Why this is useful_
This aligns the "More" button behavior with standard UI patterns found elsewhere in the Spring Initializr interface, providing a more predictable and fluid experience for developers interacting with the Share and Bookmark features.

### DIFF
--- a/USING.adoc
+++ b/USING.adoc
@@ -143,7 +143,7 @@ easily. If you are writing a tutorial that requires certain dependencies to be s
 you can also generate a link that you can use as a reference.
 
 To use this feature, simply use the UI like you would do to create a new project. Once
-you're done, click the "Share" button. A pop-up opens with a generated link with a handy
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks @aalahyan3 !

PR #2185, opened 2026-05-12
Remove extra logic for grpc test starter

dsyer:
The Boot issue that the old code was working around is fixed now, so we don't have to do anything special (and a project generated from the current code doesn't work).

I can't get a build to complete, but the problem seems unrelated (hangs on a random test).

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/AddTestStartersBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/AddTestStartersBuildCustomizer.java
@@ -17,7 +17,6 @@
 package io.spring.start.site.extension.dependency;
 
 import java.util.List;
-import java.util.Map;
 
 import io.spring.initializr.generator.buildsystem.Build;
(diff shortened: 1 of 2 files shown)


dsyer:
There's also a problem with the generated Gradle projects - they don't build because the protobuf plugin fails. To fix them I had to add an explicit protoc version:

```
protobuf {
	protoc {
		artifact = "com.google.protobuf:protoc:${dependencyManagement.importedProperties['protobuf-java.version']}"
	}
}
```

Not sure if that's right, or if it's still broken with kotlin, so I'll hold off making that change as well.

UPDATE: https://github.com/spring-projects/spring-boot/issues/50401 (it's only an issue with client projects)

dsyer (commented):
I also didn't check if this is only a problem for 4.1.0-SNAPSHOT (maybe the old code works for 4.1.0-RC1?). The issue says it was closed in RC1 (https://github.com/spring-projects/spring-boot/issues/49690) so it's probably fine.

mhalbritter:
Thanks @dsyer !

PR #2186, opened 2026-05-12
Update timefold-solver version to 2.1.0

zepfred:


mhalbritter:
Hey @zepfred , i'm not a fan of adding a new entry for v2 of timefold solver. We can give users, when they select Spring Boot 4, v2 of timefold solver without adding a new entry. This is what you want to do, right?

zepfred:
> Hey @zepfred , i'm not a fan of adding a new entry for v2 of timefold solver. We can give users, when they select Spring Boot 4, v2 of timefold solver without adding a new entry. This is what you want to do, right?

Hey @mhalbritter, Timefold solver version `1.x` only supports Spring Boot `3.5.x` and requires Java 17, whereas Timefold `2.x` requires Spring Boot `4.x` and Java 21. The goal of adding two entries was to include a description of the Java version for the Timefold `2.x` dependency. I couldn't find a way to enforce the Java version, and I'm okay with reverting the change to the previous approach using a single dependency instead.

mhalbritter:
I've reduced it back to one entry. You can see the forced Java 21 upgrade in this commit: https://github.com/spring-io/start.spring.io/pull/2186/changes/46f7676d176e175cb204951658345eb0c388cfc0

mhalbritter:
Thanks @zepfred !

PR #2203, opened 2026-06-03
Add entry for Tanzu Spring SDK Enterprise

csterwa:
Added Tanzu Spring SDK as an Enterprise labeled dependency that can be added to Spring Boot 3.5+ applications.


VERIFICATION_EXAMPLES.md
```diff
@@ -0,0 +1,91 @@
...
+## Configuration Added to start.spring.io
+
+### 1. BOM Definition (application.yml)
+```yaml
+tanzu-spring-sdk:
+  groupId: com.vmware.tanzu.spring
+  artifactId: tanzu-spring-sdk-dependencies
+  versionProperty: tanzu-spring-sdk.version
+  mappings:
+    - compatibilityRange: "[3.5.0,4.0.0)"
```
csterwa: @mhalbritter I think that this is incorrect since this should work for 3.5+. Is that correct?

mhalbritter: I've updated the range. It now supports 3.5.x and 4.0.x. That's correct, isn't it?

mhalbritter:
Thanks @csterwa !

PR #2204, opened 2026-06-03
Make zipkin build configuration more minimal on Spring Boot 4.0 or later

### LINKED ISSUE #2196: Make zipkin build configuration more minimal on Spring Boot 4.0 or later
I have the impression that the zipkin starter could be a bit more minimal on Spring Boot >= 4.0. I will explain why.

At the moment, if you add `Zipkin` to the intializer you will get:

```kotlin
  implementation("org.springframework.boot:spring-boot-micrometer-tracing-brave")
  implementation("org.springframework.boot:spring-boot-starter-actuator")
  implementation("org.springframework.boot:spring-boot-starter-zipkin")
  implementation("io.micrometer:micrometer-tracing-bridge-brave")
  testImplementation("org.springframework.boot:spring-boot-micrometer-tracing-test")
  testImplementation("org.springframework.boot:spring-boot-starter-zipkin-test")
```

The current Spring Boot reference [documentation](https://docs.spring.io/spring-boot/reference/actuator/tracing.html#actuator.micrometer-tracing.tracer-implementations.brave-zipkin), states: 

> Tracing with OpenZipkin Brave and reporting to Zipkin requires the following dependencies:
org.springframework.boot:spring-boot-starter-zipkin

In addition, the spring-boot-starter-zipkin is already pulling these two dependencies, according to its dependency [tree](https://central.sonatype.com/artifact/org.springframework.boot/spring-boot-starter-zipkin/4.0.6).

```xml
<!-- pom.xml: spring-boot-starter-zipkin -->
<dependency>
      <groupId>org.springframework.boot</groupId>
      <artifactId>spring-boot-micrometer-tracing-brave</artifactId>
      <version>4.0.6</version>
      <scope>compile</scope>
</dependency>

<dependency>
      <groupId>io.micrometer</groupId>
      <artifactId>micrometer-tracing-bridge-brave</artifactId>
      <version>1.6.5</version>
      <scope>runtime</scope>
      ...
</dependency>
```

My idea would be to remove the two redundant dependencies since the `spring-boot-starter-zipkin` has been added for convenience on Spring Boot 4.0.

Similarly, adding `OpenTelemetry` to the initializer only pulls the `spring-boot-starter-opentelemetry` to your application, and its `spring-boot-starter-opentelemetry-test` dependency companion.

If you agree that the starter in `4.x` resolves the need for these two dependencies, I can work on dropping them on the condition of Spring Boot being on >= `4.0`.

The `spring-boot-starter-zipkin-test` also [depends](https://central.sonatype.com/artifact/org.springframework.boot/spring-boot-starter-zipkin-test/dependencies) on `spring-boot-micrometer-tracing-test`, which would be the test dependency to clean up from this setup.

It might also be that I'm missing something obvious, please correct me if my thinking is wrong.

albihasani94:
We should use the minimal set of dependencies that work for the given purpose around observability. This PR resolves spring-io/start.spring.io#2196

### zipkin

> Enable and expose span and trace IDs to Zipkin.

On Spring Boot 4.0, corresponds to `org.springframework.boot:spring-boot-starter-zipkin`
On Spring Boot 3.5, corresponds to `io.zipkin.reporter2:zipkin-reporter-brave`

### opentelemetry

> Publish metrics and traces in OpenTelemetry's OTLP format.

On Spring Boot 4.0, corresponds to `org.springframework.boot:spring-boot-starter-opentelemetry`
There is no Spring-owned started on Spring Boot 3.5.

### distributed-tracing

> Enable span and trace IDs in logs.

On Spring Boot 4.0, corresponds to the following dependencies:

```
org.springframework.boot:spring-boot-micrometer-tracing-brave
io.micrometer:micrometer-tracing-bridge-brave
```

On Spring Boot 3.5, corresponds to `io.micrometer:micrometer-tracing-bridge-brave`

## Current situation

- On Spring Boot 4, selecting `zipkin` will also add `distributed-tracing` to the generated project.
  - The zipkin starter already includes the Brave dependencies, making this a matter of explicitness only
- On Spring Boot 4, selecting either `zipkin` or `opentelemetry` in combination with `distributed-tracing` will also add the full `distributed-tracing` dependencies to the generated project.
  - In case of zipkin, this seems like simple duplication of Brave instrumentation. in case of otel, the Brave dependencies do not have to be there at all

## Goals
- On Spring Boot 4, Zipkin should not add distributed-tracing.
  - Opentelemetry already does not
- On Spring Boot 3.5, zipkin should continue to come with distributed-tracing
- If either Zipkin or Otel starter is present in combination with distributed-tracing, remove the redundant tracing implementation coming from distributed-tracing
  - Zipkin already includes Brave instrumentation to produce trace and span ids
  - Otel have their own tracer which also contributes trace and span ids to the logs

## Testing

I have started by making new tests first, letting them fail on the existing implementation, then tweak the project generation and customization in order to pass the updated tests.

I would like to take some more time to do some runtime testing with the generated projects. That means testing:

- [ ] a project with Zipkin on Spring Boot 4 (We made the  change of removing distributed-tracing)
- [ ] a project with Otel + distributed-tracing on Spring Boot 4 (This will be same as Otel only, we removed distributed-tracing for the reasons mentioned above)

## Examples

On the current implementation, adding `zipkin`:

```
dependencies {
  implementation("org.springframework.boot:spring-boot-micrometer-tracing-brave")
  implementation("org.springframework.boot:spring-boot-starter-actuator")
  implementation("org.springframework.boot:spring-boot-starter-zipkin")
  implementation("io.micrometer:micrometer-tracing-bridge-brave")
  testImplementation("org.springframework.boot:spring-boot-micrometer-tracing-test")
  testImplementation("org.springframework.boot:spring-boot-starter-zipkin-test")
  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}
```

Adding `zipkin` on the updated implementation:

```
dependencies {
  implementation("org.springframework.boot:spring-boot-starter-actuator")
  implementation("org.springframework.boot:spring-boot-starter-zipkin")
  testImplementation("org.springframework.boot:spring-boot-starter-zipkin-test")
  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}
```

On the current implementation, adding `opentelemetry` + `distributed-tracing`:

```
dependencies {
  implementation("org.springframework.boot:spring-boot-micrometer-tracing-brave")
  implementation("org.springframework.boot:spring-boot-starter-actuator")
  implementation("org.springframework.boot:spring-boot-starter-opentelemetry")
  implementation("io.micrometer:micrometer-tracing-bridge-brave")
  testImplementation("org.springframework.boot:spring-boot-micrometer-tracing-test")
  testImplementation("org.springframework.boot:spring-boot-starter-opentelemetry-test")
  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}
```

Adding `opentelemetry` + `distributed-tracing` on the updated implementation:

```
dependencies {
  implementation("org.springframework.boot:spring-boot-starter-actuator")
  implementation("org.springframework.boot:spring-boot-starter-opentelemetry")
  testImplementation("org.springframework.boot:spring-boot-starter-opentelemetry-test")
  testRuntimeOnly("org.junit.platform:junit-platform-launcher")
}
```

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/observability/ObservabilityActuatorBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/observability/ObservabilityActuatorBuildCustomizer.java
@@ -63,13 +63,18 @@ class ObservabilityActuatorBuildCustomizer implements BuildCustomizer<Build> {
 		}
 	}
 
+	@Override
+	public int getOrder() {
+		return ObservabilityDistributedTracingBuildCustomizer.ORDER + 1;
+	}
+
 	private boolean hasActuator(Build build) {
 		return build.dependencies().has("actuator");
 	}
 
 	private boolean needsActuator(Build build) {
 		if (isBoot4OrLater()) {
-			return hasTracing(build) || hasPullBasedMetrics(build);
+			return hasPullBasedMetrics(build);
 		}
 		else {
 			return hasTracing(build) || hasPullBasedMetrics(build) || hasPushBasedMetrics(build);
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/observability/ObservabilityDistributedTracingBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/observability/ObservabilityDistributedTracingBuildCustomizer.java
@@ -32,6 +32,8 @@ import io.spring.initializr.generator.version.VersionRange;
  */
 class ObservabilityDistributedTracingBuildCustomizer implements BuildCustomizer<Build> {
 
+	static final int ORDER = 0;
+
 	private static final VersionRange SPRING_BOOT_4_OR_LATER = VersionParser.DEFAULT.parseRange("4.0.0-M1");
 
 	private static final String DISTRIBUTED_TRACING = "distributed-tracing";
@@ -45,20 +47,23 @@ class ObservabilityDistributedTracingBuildCustomizer implements BuildCustomizer<
 	@Override
 	public void customize(Build build) {
 		if (isBoot4OrLater()) {
-			if (hasTracingProvider(build)) {
-				build.dependencies().remove(DISTRIBUTED_TRACING);
-			}
+			handleBoot4(build);
 		}
-		else if (build.dependencies().has("zipkin") && !build.dependencies().has(DISTRIBUTED_TRACING)) {
-			build.dependencies().add(DISTRIBUTED_TRACING);
+		else {
+			handleBoot35(build);
 		}
-		if (build.dependencies().has("wavefront") && build.dependencies().has(DISTRIBUTED_TRACING)) {
-			build.dependencies()
-				.add("wavefront-tracing-reporter",
-						Dependency.withCoordinates("io.micrometer", "micrometer-tracing-reporter-wavefront")
-							.scope(DependencyScope.RUNTIME));
+	}
+
+	@Override
+	public int getOrder() {
+		return ORDER;
+	}
+
+	private void handleBoot4(Build build) {
+		if (hasTracingProvider(build)) {
+			build.dependencies().remove(DISTRIBUTED_TRACING);
 		}
-		if (build.dependencies().has(DISTRIBUTED_TRACING) && isBoot4OrLater()) {
+		else if (hasDistributedTracing(build)) {
 			build.dependencies()
 				.add("spring-boot-micrometer-tracing-brave",
 						Dependency.withCoordinates("org.springframework.boot", "spring-boot-micrometer-tracing-brave"));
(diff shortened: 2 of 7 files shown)


albihasani94:
I noticed a case that is slightly different around `actuator`.

We expect selecting `Otel` + `distributed-tracing` to bring only otel dependencies. Due to `ObservabilityActuatorBuildCustomizer` temporarily finding `distributed-tracing` before `ObservabilityDistributedTracingBuildCustomizer` removes it, we end up with a harmless, but probably not explicitly needed actuator dependency. 

I think I would not touch with the order of customizers unless it was really important. In this case, I just want to expand my understanding behind the reasoning.

I read that we evaluate `hasTracing(build)` for both before and after Spring Boot 4 configurations.

```java
    private boolean needsActuator(Build build) {
		if (isBoot4OrLater()) {
			return hasTracing(build) || hasPullBasedMetrics(build);
		}
		else {
			return hasTracing(build) || hasPullBasedMetrics(build) || hasPushBasedMetrics(build);
		}
	}
```

For Zipkin, it is already there in any case, because of:

```java
Set<String> TRACING = Set.of("distributed-tracing", "zipkin")
```

Does the `TRACING` set also need to contain `opentelemetry` for consistency or real necessity? In the real-world, does Zipkin still need an explicit actuator dependency on Spring Boot 4, and how does Otel seem to do fine without it?

https://docs.spring.io/spring-boot/4.0/reference/actuator/tracing.html

> Spring Boot Actuator provides dependency management and auto-configuration for Micrometer Tracing, a facade for popular tracer libraries.

I will spend some time myself to try to get the answers.

In production, this is rarely a case for confusion as `actuator` could be there anyway, e.g. for prometheus

mhalbritter:
> In the real-world, does Zipkin still need an explicit actuator dependency on Spring Boot 4, and how does Otel seem to do fine without it?

Good catch. Neither opentelemetry, nor zipkin, nor distributed-tracing need actuator on the classpath to work with Boot 4.0+. I'll handle that in the polish commit.

mhalbritter:
Thanks a lot for the PR, @albihasani94 !

PR #2250, opened 2026-08-04
Maven projects contain unneeded exclusions for annotation processors

mvitz:
Because dependencies marked with `<optional>true</optional>` within Maven are now excluded by default there is no need to configure the explicit exclusion for `lombok` because it's always generated as an optional dependency.


mhalbritter:
Thanks for the PR, @mvitz !

Actually, we can remove `AnnotationProcessorExclusionBuildCustomizer` altogether, as every annotation processor is `optional = true` and Boot 4.0+ excludes those by default. I'll handle this in a polish commit.

PR #2261, opened 2026-08-20
Upgrade to Spring Cloud 2025.1.3

spring-builds:
Generated by the [post release workflow](https://github.com/spring-cloud/spring-cloud-github-actions/actions/runs/32407476372).

Spring Cloud 2025.1.3 has been released, so the `spring-cloud` bom mapping for the 2025.1 line moves from `2025.1.2` to `2025.1.3`. The `compatibilityRange` is unchanged.

### DIFF
--- a/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractGradleBuildCustomizer.java
+++ b/start-site/src/main/java/io/spring/start/site/extension/dependency/springcloud/SpringCloudContractGradleBuildCustomizer.java
@@ -48,6 +48,15 @@ abstract class SpringCloudContractGradleBuildCustomizer implements BuildCustomiz
 		.snapshotsEnabled(true)
 		.build();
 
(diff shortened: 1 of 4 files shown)


mhalbritter:
Spring Cloud 2025.1.3 removes the dependency management for Spring Cloud Contract, this is why the PR is failing.

mhalbritter:
I've hardcoded Spring Cloud Contract to 5.0.3 and verified that it works with Spring Cloud 2025.1.3. The migration to Stubborn Contract is tracked in https://github.com/spring-io/start.spring.io/issues/2264.

ryanjbaxter:
Thanks @mhalbritter!  I was going to look into this on Tuesday.  I saw the test was using Contract so I figured that was the problem.  I don't think we need to migrate to Stubbornish.  Instead I think we can remove the test entirely or use a different project in the test (I am not sure exactly what the test was doing since I didn't have a chance to look into it)

PR #2266, opened 2026-08-25
Provide ojdbc17 instead of ojdbc11

wimdeblauwe:
Similar to #1366 , since we are on JDK 17 at minimum, we should use the `ojdbc17` instead of `ojdbc11` variant by default I think.  This is the Oracle description for `ojdbc17`:

> Supports JDBC 4.3 spec and for use with JDK21, JDK19, and JDK17. Compatible with the Jakarta APIs. (Starting from 23.6)


mhalbritter:
That makes sense. Thanks @wimdeblauwe !

PR #2279, opened 2026-09-08
Fix LDAP starter scope for embedded LDAP

### LINKED ISSUE #2271: Adding Spring LDAP and Embedded LDAP server adds spring-boot-starter-ldap twice
One dependency with no scope, and one with scope test - when using Maven, this results in classes from spring-boot-starter-ldap not available on the compile classpath.

AtharvaBarve:
Fixes gh-2271

**Description**

When both Spring LDAP and Embedded LDAP Server are selected, `spring-boot-starter-ldap` is generated with the `test` scope. This means the LDAP starter is not available on the compile classpath, even though it is required for the application's LDAP configuration.

This change removes the `test` scope from `spring-boot-starter-ldap` while keeping `unboundid-ldapsdk` test-scoped.

This is my first contribution to `start.spring.io`, so I may have missed something or approached the fix differently than expected. I'd really appreciate any feedback or suggestions for improvement.


mhalbritter:
Thanks for the PR, but this is not the fix i was looking for: While it fixes the reported symptom (scope conflict), it doesn't fix the root cause: selecting both "LDAP" and "Embedded LDAP Server" still declares `spring-boot-starter-ldap` twice in the generated pom — just with matching scope now instead of conflicting scope. No test reproduces the actual bug (combining both dependencies); the updated test only exercises "Embedded LDAP Server" alone.

PR #2283, opened 2026-09-13
Fix project type restoration from history and bookmarks

### LINKED ISSUE #2282: Selecting a project from History or Bookmarks fails to restore Project Type due to hash truncation
### Description

When restoring a project configuration from the **History** modal or the **Bookmarks (Favorites)** modal, the project type (build system: Maven / Gradle - Groovy / Gradle - Kotlin) is not restored. It silently retains whichever project type is currently selected.

### Cause & Root Cause Analysis

1. In `start-client/src/components/common/history/Modal.js:159` and `start-client/src/components/common/favorite/Modal.js:21`, the links are rendered with a single `#`:
   ```jsx
   <a className='item' href={`/#${value}`} onClick={() => { onClose() }}>
   ```
   where `value` is formatted like `type=gradle-project&language=java...`.

2. When clicked, `window.location.hash` becomes `#type=gradle-project&...`.

3. However, in `start-client/src/components/utils/Hash.js:41`:
   ```javascript
   useEffect(() => {
     if (complete && hash) {
       const params = queryString.parse(`?${hash.substr(2)}`)
       dispatch({ type: 'LOAD', payload: { params, lists: config.lists } })
       ...
   ```
   `useHash` blindly runs `hash.substr(2)`, assuming a 2-character prefix `#!` (which is only used by the Share modal `Popover.js`).

4. Running `hash.substr(2)` on `#type=gradle-project` drops `#t`, resulting in `ype=gradle-project`:
   ```javascript
   queryString.parse('?ype=gradle-project&language=java')
   // => { ype: 'gradle-project', language: 'java' }
   ```

5. When passed to `parseParams` in `ApiUtils.js`, `'ype'` does not match `PROPERTIES_MAPPING_URL` (which expects `'type'`), so the `project` (`type`) attribute is completely ignored.

Additionally, line 18 in `Hash.js` contains a minor bug in the fallback path:
```javascript
window.history.hash = ``
```
which should be `window.location.hash = ''` as `window.history` does not have a `hash` property.

### Steps to Reproduce

1. Open https://start.spring.io/
2. Change the Project type to **Gradle - Kotlin** (or **Gradle - Groovy**).
3. Click **Generate** (to save to History) or click **Explore** -> **Add to Bookmarks** (`B`).
4. Switch the Project type back to **Maven**.
5. Open the **History** modal (`H`) or **Bookmarks** modal (`B`) and click on the previously saved entry.
6. **Observed**: The language, dependencies, and metadata update, but Project remains **Maven**.
7. **Expected**: Project type should switch to **Gradle - Kotlin**.

### Proposed Fix

In `start-client/src/components/utils/Hash.js`:
Strip any leading `#` or `#!` before parsing query parameters:
```javascript
const hashString = hash.replace(/^#[!/]?/, '')
const params = queryString.parse(`?${hashString}`)
```
And update `clearHash`:
```javascript
window.location.hash = ''
```

---

Can I please be assigned to this issue? I'd be happy to submit a pull request with the fix and accompanying tests.

Gaalbu:
Fixes #2282.

When a configuration was restored from History or Bookmarks, the hash contained a single `#`, but the parser always removed two characters as if the hash used the `#!` share URL format. This dropped the first character of `type`, so the selected project type was not restored.

This change:

- strips the optional `#!` or `/` prefix while preserving hashes with a single `#`;
- fixes the fallback hash clearing path to use `window.location.hash`;
- adds regression coverage for both hash formats.

Validation:

- `yarn test --runInBand` (55 tests passing)
- targeted ESLint on changed files
- production build
- `npx aislop scan --changes --json` (100/100)


Gaalbu:
Friendly ping — this PR is still waiting for triage. Let me know if any changes are needed.

mhalbritter:
Hello @Gaalbu , the submitter of the bug asked to fix it:

> Can I please be assigned to this issue? I'd be happy to submit a pull request with the fix and accompanying tests.

Thanks for the PR anyway.

